PR #10, opened 2022-01-08
Copy github standard Visual Studio gitignore

BraydonKains:
Using the GitHub standard Visual Studio gitignore, see [original](https://github.com/github/gitignore/blob/main/VisualStudio.gitignore) for reference.


zysim:
LGTM

PR #18, opened 2022-01-22
Add Json Serialization Option

### LINKED ISSUE #17: Task: Return DTOs to Frontend as camelCase
The DTO the frontend receives is in PascalCase, allegedly. We should make it so that it's in camelCase cuz that's cooler :sunglasses: 

### Affected Functionality

Only on the frontend.

### Other Relevant Issues

N/A

RageCage64:
Closes #17 

Add a json serialization option to camelcase all incoming and outgoing json serialized and deserialized by the controllers.

### DIFF
--- a/LeaderboardBackend/Program.cs
+++ b/LeaderboardBackend/Program.cs
@@ -5,6 +5,7 @@ using Microsoft.IdentityModel.Tokens;
 using System.Text;
 using LeaderboardBackend.Services;
 using System.IdentityModel.Tokens.Jwt;
+using System.Text.Json;
 
 JwtSecurityTokenHandler.DefaultInboundClaimTypeMap.Clear();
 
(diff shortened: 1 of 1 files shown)


zysim:
Would we need to change anything in controller code with this change?

zysim:
Would we need to call `services.AddMvc().AddJsonOptions()` too?

RageCage64:
>Would we need to change anything in controller code with this change?
Nope. It uses this serialization automatically once the controllers are added to the app.

>Would we need to call services.AddMvc().AddJsonOptions() too?
Nope. This app does not use MVC, since it is only an API. If you wanted to add views for whatever reason in the future, then you would need to add it there too.

zysim:
Oh yeah, true, this isn't an MVC project lmao. Aight sweet I'll fix the merge conflict and merge this.

zysim (approved):

Issue #5, opened 2022-01-06
Add remaining unit tests


RageCage64:
The codebase currently needs to catch up in terms of test case coverage, so that all future features have a good base of tests to work with.

Note: From here on out, all new enhancements will imply that tests follow, this should be the only time tests need their own issue.

RageCage64:
The only set of tests left is UserService

zysim:
> The only set of tests left is UserService
After reading the [docs on writing tests with the in-memory DB](https://docs.microsoft.com/en-us/ef/core/testing/#approach-3-the-ef-core-in-memory-database), it sounds like we should hold off on writing tests for this until #23 gets merged, on which we should write tests that target the test Postgres DB, right?

PR #19, opened 2022-01-22
Added AuthService test for GenerateJSONWebToken

RageCage64:
Work for #5 

Added AuthService test for GenerateJSONWebToken


.gitignore
```diff
@@ -378,9 +378,9 @@ FodyWeavers.xsd
 
 # VS Code files for those working on multiple tools
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
+.vscode/settings.json
```
RageCage64, line 381: Meant for these to be ignored, accidentally left the `!` in

LeaderboardBackend/Services/UserService.cs
```diff
@@ -1,4 +1,4 @@
-using System.IdentityModel.Tokens.Jwt;
+using System.IdentityModel.Tokens.Jwt;
```
RageCage64, line 1: line ending change for reasons 🤷 
I guess I committed it with the wrong one before

leaderboard-backend-poc.sln
```diff
@@ -1,35 +1,35 @@
...
-	EndGlobalSection
-	GlobalSection(SolutionProperties) = preSolution
-		HideSolutionNode = FALSE
-	EndGlobalSection
-	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {FE37FED9-098E-4D35-9608-99D85C7B36A2}
-	EndGlobalSection
-EndGlobal
\ No newline at end of file
+Microsoft Visual Studio Solution File, Format Version 12.00
```
RageCage64, line 1: My suspicion is that Visual Studio does not respect the editorconfig with the `sln` file. So if we commit this and leave it, then Visual Studio should stop being upset by the wrong format. If this doesn't solve it, we might need to edit our editorconfig file to match was Visual Studio does to this file by default, so opening it in a different editor doesn't make the whole thing go berserk lmao

.gitignore
zysim: Shouldn't the line above cover this ignore then?

LeaderboardBackend.Test/Services/AuthServiceTests.cs
```diff
@@ -0,0 +1,58 @@
...
+		string key = "testkeythatsatisfiesthecharacterminimum";
+		string issuer = "leaderboards.gg";
+
+		string configJson = string.Format(@"
+		{{
+			""Jwt"": {{
+				""Key"": ""{0}"",
+				""Issuer"": ""{1}""
+			}}
+		}}
```
zysim, line 25: We could try making the keys camelCase here maybe?

leaderboard-backend-poc.sln
zysim: If someone else could verify this works pls :+1:

LeaderboardBackend.Test/Services/AuthServiceTests.cs
RageCage64: I'm matching it exactly to how `appsettings.json` does it, and in the AuthService they are accessed with their pascal cased names.

zysim: Ah, okay then.

leaderboard-backend-poc.sln
zysim: I'll just go out on a limb and say that this works. If anyone else who uses VS and still gets this error, we'll deal with it in the future. I'll make an issue to reference this just in case.

zysim (approved):
Tests pass, and the code looks good. Merging.

Issue #4, opened 2022-01-06
Add Postgres support


RageCage64:
As of writing, the application is using an InMemory database. This is useful for quick development and it would be great to leave it in, but we also want to add real Postgres support.

Requirements:
* The app will need `.env` support, suggested https://www.nuget.org/packages/dotenv.net/
* A `docker-compose.yml` file that will run the `postgres` and `adminer` containers (see https://github.com/leaderboardsgg/leaderboard-backend-go/blob/main/docker-compose.yml)
* An `example.env` file that developers can copy for their local `.env` (gitignored) file to run the database and get credentials in the app
* Add all new relevant instructions to the setup directions located in the README.md file

PR #23, opened 2022-01-23
[REVIEW FIRST] Postgres Support

zysim:
Closes: #4 

Adds:
- Reading things from `.env`
- Registering `DbContext`s to use Postgres

Note: Our Postgres config is loaded via `.env` instead of using secrets or JSONs. This is because `docker-compose.yml` reads from `.env` by default, so we'd might as well source that same file for our config.


.editorconfig
```diff
@@ -7,3 +7,8 @@ end_of_line = lf
 insert_final_newline = true
 indent_style = tab
 charset = utf-8
+
+# docker-compose.yml doesn't like tabs lmao
+[.yml]
+indent_style = space
+indent_size = 4
```
RageCage64, line 14: Some people want yaml to be 2 spaces, I don't think it matters much to me though and it's easy enough to change

RageCage64 (approved):
Your changes all look good to me @zysim so if you're good with mine, we can merge this in.

.editorconfig
zysim: ¯\\_(ツ)\_/¯ if anyone else wants to change this really inconsequential thing later in the future they can. It's not worth tacking on another commit just to change a number.

zysim:
Fixed a type cast. Merging this in. :tada:

Issue #20, opened 2022-01-22
Enforce Validation On Model Creation


zysim:
Paraphrasing the referenced comment (_italics_ mine):

> We should start adding validation to the models before we create them. We could call some `leaderboard.Validate()` method before creating them in _the corresponding service_

_Originally posted by @RageCage64 in https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/14#discussion_r790097417_

PR #29, opened 2022-01-27
MVP: User Model Validation

zysim:
Closes: #20.
This should serve as a foundation for other model validations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -301,7 +301,7 @@ node_modules/
 *.dsw
 *.dsp
 
-# Visual Studio 6 technical files 
+# Visual Studio 6 technical files
 *.ncb
 *.aps
 
@@ -378,9 +378,7 @@ FodyWeavers.xsd
 
 # VS Code files for those working on multiple tools
 .vscode/*
-.vscode/settings.json
-.vscode/tasks.json
-.vscode/launch.json
+!.vscode/tasks.json
 !.vscode/extensions.json
 *.code-workspace
 
--- a/LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
+++ b/LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
@@ -29,7 +29,7 @@ namespace LeaderboardBackend.Controllers.Requests
 
 		protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
 		{
-			var user = (RegisterRequest)validationContext.ObjectInstance;
+			var user = (LoginRequest)validationContext.ObjectInstance;
 			var password = user.Password!;
 
 			// Validate length
--- a/LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
+++ b/LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
@@ -5,7 +5,18 @@ namespace LeaderboardBackend.Controllers.Requests
 {
 	public class PasswordAttribute : ValidationAttribute
 	{
-		public string GetErrorMessage(List<string> missing)
+		private static readonly int MIN = 8;
+		private static readonly int MAX = 80;
+
+		public string GetErrorMessageLength(bool tooShort)
+		{
+			if (tooShort) {
+				return $"Your password needs to be at least {MIN} characters long.";
+			}
+			return $"Your password needs to be at most {MAX} characters long.";
+		}
+
+		public string GetErrorMessageMissing(List<string> missing)
 		{
 			if (missing.Count == 1) {
 				return $"Your password still needs {missing[0]}.";
@@ -21,6 +32,14 @@ namespace LeaderboardBackend.Controllers.Requests
 			var user = (RegisterRequest)validationContext.ObjectInstance;
 			var password = user.Password!;
 
+			// Validate length
+			if (password.Length < MIN) {
+				return new ValidationResult(GetErrorMessageLength(true));
+			} else if (password.Length > MAX) {
+				return new ValidationResult(GetErrorMessageLength(false));
+			}
+
+			// Validate content
 			var missing = new List<string>();
 
 			if (!new Regex(@"[a-z]").IsMatch(password)) {
@@ -37,7 +56,7 @@ namespace LeaderboardBackend.Controllers.Requests
 
 			if (missing.Count > 0)
 			{
-				return new ValidationResult(GetErrorMessage(missing));
+				return new ValidationResult(GetErrorMessageMissing(missing));
 			}
 
 			return ValidationResult.Success;
--- a/LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
+++ b/LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
@@ -13,10 +13,10 @@ namespace LeaderboardBackend.Controllers.Requests
 
 		protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
 		{
-			var user = (LoginRequest)validationContext.ObjectInstance;
-			var password = user.Password!;
+			LoginRequest user = (LoginRequest)validationContext.ObjectInstance;
+			string password = user.Password;
 
-			var errors = new List<string>();
+			List<string> errors = new();
 
 			if (password.Length < MIN)
 			{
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -378,9 +378,9 @@ FodyWeavers.xsd
 
 # VS Code files for those working on multiple tools
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
+.vscode/settings.json
+.vscode/tasks.json
+.vscode/launch.json
```
FrozenSake, line 383: Weren't these carve-outs especially to include those? If we want to not include them, we should delete the line because the `.vscode/*` line already catches them.

FrozenSake (commented):
Most stuff looks good to me, but I'm not great with C#, so I stuck to one thing I know 😏

.gitignore
zysim: Yeah I really don't know honestly. In fact I think we should track `.vscode/{extensions,launch}.json`, as there are custom values in there. That or track a `.vscode/{extensions,launch}.json.example` at least

FrozenSake: They're *not* ignored currently. A `!` at the start of a gitignore line says "include this". So with
```
.vscode/*
!.vscode/launch.json
```
it will ignore the entire .vscode folder, except launch.json.

zysim: It was because we originally didn't have anything substantial in those files. But now there's `launch.json` and `extensions.json` that will be good to track. I'll be editing these lines in `.gitignore` to track those two files in the folder.

zysim: https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/29/commits/f20ea6ec8fcfd572dcb41969f7552c61966694ac

FrozenSake: 👍 The new commit has us only removing the inclusion on settings and tasks, which seems better.

zysim:
I think I'll stop work on this PR here. There's gonna be a _lot_ of tests to work over, but for the sake of not ballooning this PR up further, I'll delegate that to another issue.

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		public string GetErrorMessageLength(bool tooShort)
+		{
+			if (tooShort) {
+				return $"Your password needs to be at least {MIN} characters long.";
+			}
+			return $"Your password needs to be at most {MAX} characters long.";
+		}
+
+		public string GetErrorMessageMissing(List<string> missing)
```
RageCage64, line 19: I think it would be a lot cleaner to just come up with one sentence where it doesn't matter how many errors follow. Maybe something like 
`Your password has the following errors: <all the errors>`

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Text.RegularExpressions;
+
+namespace LeaderboardBackend.Controllers.Requests
+{
+	public class PasswordAttribute : ValidationAttribute
+	{
+		private static readonly int MIN = 8;
+		private static readonly int MAX = 80;
+
+		public string GetErrorMessageLength(bool tooShort)
```
RageCage64, line 11: These methods are only used in this class in one place. I think either this methods should be made private and put at the bottom, or they should just be inlined to the `IsValid` method. Up to you, if it were me I'd inline them and use the simplification from the other comment, but I'll let you decide.

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+			}
+			if (missing.Count == 2) {
+				return $"Your password still needs {missing[0]} and {missing[1]}.";
+			}
+			return $"Your password still needs {String.Join(", ", missing.GetRange(0, missing.Count - 1).ToArray())}, and {missing[missing.Count - 1]}";
+		}
+
+		protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
+		{
+			var user = (LoginRequest)validationContext.ObjectInstance;
```
RageCage64, line 32: This PR has existed longer than this decision so obviously you couldn't have known for this PR, but I think we're moving away from using `var` at all.

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+
+			// Validate length
+			if (password.Length < MIN) {
+				return new ValidationResult(GetErrorMessageLength(true));
+			} else if (password.Length > MAX) {
+				return new ValidationResult(GetErrorMessageLength(false));
+			}
+
+			// Validate content
+			var missing = new List<string>();
```
RageCage64, line 43: So for this with the move away from `var` you can instead go 
`List<string> missing = new();`

LeaderboardBackend/Controllers/Requests/Register.cs
```diff
@@ -1,9 +1,14 @@
...
 {
+	[Required]
+	[RegularExpression("/([a-zA-Z][-_']?){1,12}[a-zA-Z]/",
+		ErrorMessage = "Your name must be between 2 and 25 characters, made up of letters sandwiching zero or one hyphen, underscore, or apostrophe.")]
 	public string Username { get; set; } = null!;
-	public string Email { get; set; } = null!;
-	public string Password { get; set; } = null!;
+
+	[Required]
+	[Compare("Password", ErrorMessage = "This must match your password.")]
```
RageCage64, line 12: I think this should say "The password confirmation must match you password" instead of "this".

LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -36,11 +36,22 @@ public async Task<ActionResult<User>> GetUser(Guid id)
...
 	{
+		// This shouldn't hit normally, since we have the CompareAttribute in Register.cs
 		if (body.Password != body.PasswordConfirm)
 		{
 			return BadRequest();
 		}
 
+		if (await _userService.GetUserByEmail(body.Email) != null)
+		{
+			return Conflict("A user already exists with this email.");
```
RageCage64, line 47: So let's leave this here, but I think the plan overall is not to use `409 Conflict` here worth keeping in mind

LeaderboardBackend/Models/User.cs
```diff
@@ -1,11 +1,12 @@
...
 
 namespace LeaderboardBackend.Models;
 
 public class User
 {
 	public Guid Id { get; set; }
 	[Required] public string Username { get; set; } = null!;
-	[Required] public string Email { get; set; } = null!;
-	[Required] public string Password { get; set; } = null!;
+	[Required] [EmailAddress] public string Email { get; set; } = null!;
```
RageCage64, line 10: I think it's nicer for each annotation tag to have its own line, similar to `LoginRequest` in this PR. I think it's more consistent, since a lot of properties are going to have more than one so it makes the whole thing look more aligned and keeps lines a bit shorter. I'll leave it up to you.

RageCage64 (changes requested):
One functionality change, the rest are just style things and nits

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
zysim: Yeah I'm just consolidating this into a `GetErrorMessage` method.

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
zysim: Oh yeah. Gonna update everything.

LeaderboardBackend/Controllers/Requests/PasswordAttribute.cs
zysim: Ye; learned of that sugar since.

LeaderboardBackend/Controllers/UsersController.cs
zysim: Tru. I'll add a comment. In fact; the `MiddlewareResultHandler` I've written in [the authZ PR](https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/60/files#diff-a1f7e3047926dcd210c9097f55ef957ce5234adf163f98e2d53cf4d9570ccea8R30-R32) should be able to override this, if I'm not mistaken.

LeaderboardBackend/Models/User.cs
zysim: Yeah I was trying to emulate what Ero's done with his changes. I think I'll do an attribute on a line if there are multiple attributes for a field.

RageCage64 (approved):

Issue #30, opened 2022-01-29
MVP: Complete Remaining User Interactions


zysim:
We've gotta finish up these interactions for Users:
- [ ] An API client can retrieve the name of a user, their about text, and their admin status by ID or by name. (https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/43)
- [ ] An authenticated user can retrieve their own name, email, about text, and admin status. (https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/43)
- [ ] An API client can retrieve a list of all the leaderboards that a user moderates by user ID.
- [ ] An API client can retrieve all of a user's bans by user ID.
- [ ] A user can update their own about text.
- [ ] A user can delete their own account, and doing so will remove them from any leaderboards they moderate and delete all of their runs.

Append each item with a link to the PR so we can better track progress.

### Affected Functionality
We will need to:
- Finish the remaining columns for Users: About and Admin
- Add the Modship table
- Add the relationship between Users and Modships, and Modships and Leaderboards

### Other Relevant Issues
-

PR #43, opened 2022-02-12
Complete User Fields And Interactions

zysim:
Closes: #30. Based off: #29 

- Adds last two missing fields in Users:
  - `About`; and
  - `Admin`
- Adds a migration for those two fields
- Adds controller action to get a User from their email: `GetUserByEmail`

Things to note:
- `Admin` is non-nullable and is hardcoded to `false` on registration.
- `About` is nullable and isn't added on registration.


LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -40,6 +40,24 @@ public async Task<ActionResult<User>> GetUser(Guid id)
...
 	}
 
+	/// <summary>Gets a User by email.</summary>
+	/// <param name="email">The User's email.</param>
+	/// <response code="200">The User with the provided email.</response>
+	/// <response code="404">If no User is found with the provided email.</response>
+	[ApiConventionMethod(typeof(Conventions),
+						 nameof(Conventions.Get))]
+	[HttpGet("{email}")]
+	public async Task<ActionResult<User>> GetUserByEmail(string email)
```
RageCage64, line 50: I don't think we actually want this

RageCage64 (approved):

Issue #31, opened 2022-01-29
MVP: Bans


zysim:
🔨🔨🔨

Append each item with the PR that solves it.

## Model/Relationships

- [x] Create the model - refer to [MVP](https://github.com/leaderboardsgg/leaderboard-backend-poc/wiki/MVP-Overview) (#32)
  - Zero or multiple Bans to a User
  - Zero or multiple Bans to a Leaderboard
  - Users: `BanningUser` and `BannedUser` link to a User's ID
  - Leaderboards: `Leaderboard` links to a Leaderboard's ID

## Controller Actions

- [x] GET Ban with ID (https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/53)
- [x] GET all Bans, optionally filtered by either Leaderboard ID or User ID (https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/53)
- [x] POST Bans as admin with: (https://github.com/leaderboardsgg/leaderboard-backend/pull/95)
  - User ID;
  - Reason;
  - Condition: must be admin
  - Condition: Leaderboard ID **must** be `null` i.e. admins can only give site-wide bans
  - Condition: User to be banned **must not** be admin
- [x] POST Bans as mod with: (https://github.com/leaderboardsgg/leaderboard-backend/pull/95)
  - User ID;
  - Reason;
  - Leaderboard ID;
  - Condition: Leaderboard ID **must not** be `null`
  - Condition: User to be banned **must not** be mod **nor** admin
- [ ] Soft DELETE Ban as admin with Ban ID (#101)
- [ ] Soft DELETE Ban as mod with: (#101)
  - Ban ID;
  - Leaderboard ID

## Post-action Behaviours

- [ ] New Ban for User, if mod:
  - Modship for any and all Leaderboards removed; and
  - any and all Runs across all Leaderboards rejected
- [ ] New Ban for User if not mod:
  - any and all Runs for specific Leaderboard rejected

## Other Relevant Issues
(To be filled)

TheTedder:
Bans have been retconned by #144. They're part of our rudimentary role system now.

just-ero:
Please update this issue to reflect the current status.

* deleting bans has been implemented (#101)
* users no longer have modships, which nullifies `Ban for User, if mod`

just-ero:
In fact, do we even have bans anymore? I'm looking through the code right now, and there's not a single mention of the word `ban`. What happened?

[...]

PR #53, opened 2022-02-16
MVP: Add Ban Controller

zysim:
Closes: #31. Based off: #32

Handles:
- An API client can get a list of all bans, optionally filtered by either leaderboard ID or user ID
- An API client can retrieve the details of a ban by providing an ID


RageCage64:
Covered by #79

PR #62, opened 2022-03-09
Allow Absence of Rules on POST Leaderboards

zysim:
At some point, the `Leaderboard` model required `Rules` be passed on instantiation. This PR removes that requirement, as we shouldn't have to pass rules in when creating a leaderboard.


RageCage64:
This updated use case is covered by #68

Issue #66, opened 2022-03-13
Task: Add Documentation Comments to Existing Models & Controllers


zysim:
We should add Open API doc comments to all our models and controllers. Refer to [this chapter in the ASP.NET docs](https://docs.microsoft.com/en-us/aspnet/core/tutorials/getting-started-with-swashbuckle?view=aspnetcore-6.0&tabs=netcore-cli#customize-and-extend) on how to customise and extend the generated docs from Open API.

### Affected Functionality

The generated Open API reference to our stuff will be enhanced and easier to follow.

### Other Relevant Issues

\-

PR #67, opened 2022-03-13
Open API Annotations

zysim:
Closes: #66

### DIFF
--- a/LeaderboardBackend/Controllers/Requests/Register.cs
+++ b/LeaderboardBackend/Controllers/Requests/Register.cs
@@ -1,21 +1,41 @@
 using System.ComponentModel.DataAnnotations;
 namespace LeaderboardBackend.Controllers.Requests;
 
-public record RegisterRequest 
+public record RegisterRequest
 {
+	/// <summary>User's name.</summary>
+	/// <example>Ayylmao Gaming</example>
 	[Required]
 	[RegularExpression("(?:[a-zA-Z][-_']?){1,12}[a-zA-Z]",
 		ErrorMessage = "Your name must be between 2 and 25 characters, made up of letters sandwiching zero or one hyphen, underscore, or apostrophe.")]
 	public string Username { get; set; } = null!;
 
+	/// <summary>User's email.</summary>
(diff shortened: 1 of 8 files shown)


LeaderboardBackend/Program.cs
```diff
@@ -40,6 +41,10 @@
 builder.Services.AddSwaggerGen(c =>
 {
 	c.SwaggerDoc("v1", new() { Title = "LeaderboardBackend", Version = "v1" });
+
+	// Enable adding XML comments to controllers to populate Swagger UI
+	var xmlFilename = $"{Assembly.GetExecutingAssembly().GetName().Name}.xml";
```
RageCage64, line 46:
```suggestion
	string xmlFilename = $"{Assembly.GetExecutingAssembly().GetName().Name}.xml";
```

RageCage64 (dismissed):

LeaderboardBackend/Controllers/Requests/Register.cs
```diff
@@ -1,9 +1,13 @@
+using System.ComponentModel.DataAnnotations;
+
 namespace LeaderboardBackend.Controllers.Requests;
 
-public record RegisterRequest
+public record RegisterRequest : LoginRequest
```
RageCage64, line 5: Was this on purpose? I don't think it's worth how confusing it is. We should probably keep these separate.

zysim: It is. It's because of [`PasswordAttribute.IsValid`](https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/29/files#diff-9712e1ce73def0622f8715bde6049f2b7488df631ffce0bc90a113034b20f98cR16-R17):
```csharp
LoginRequest user = (LoginRequest)validationContext.ObjectInstance;
string password = user.Password;
```
I can't resolve `user` easily.

zysim: Thanks to your changes to `main` I can edit this now :DD

RageCage64:
Closing in favour of #70 which is just a reconstructed version of this

PR #68, opened 2022-03-19
Restructured Models, added Models from all PRs

RageCage64:
This PR adds every Model currently in flight. Due to conflicts while attempting to get all the different Models in, I opted to copy the specific model from every PR, and start migrations from scratch. From now on we should not run into the issue where things are irreconcilable. Sorry @zysim this is gonna make it look like I did the models in the commit history LOL

Also restructured the Models folder to include Requests and made a central folder for Annotations, as there will almost certainly be more.


zysim:
> gonna make it look like I did the models in the commit history
Lucky for you I couldn't care any less :sunglasses:

LeaderboardBackend/Models/Entities/Ban.cs
```diff
@@ -0,0 +1,39 @@
...
+
+	/// <summary>Generated on creation.</summary>
+	[Required] 
+	public DateTime Time { get; set; } = default!;
+
+	/// <summary>ID of User who set the Ban. Must be either an admin or a mod.</summary>
+	public Guid? BanningUserId { get; set; }
+
+	[JsonIgnore] 
+	public User BanningUser { get; set; } = null!;
```
zysim, line 23: This should be nullable.

LeaderboardBackend/Models/Entities/Judgement.cs
```diff
@@ -0,0 +1,29 @@
...
+	public Guid ApproverId { get; set; }
+
+	[JsonIgnore]
+	public User? Approver { get; set; }
+
+	[Required]
+	public Guid RunId { get; set; }
+
+	[JsonIgnore]
+	public Run? Run { get; set; } = null!;
```
zysim, line 28:
```suggestion
	public Run Run { get; set; } = null!;
```

LeaderboardBackend/Models/Entities/Judgement.cs
```diff
@@ -0,0 +1,29 @@
...
+	public DateTime timestamp { get; set; }
+
+	[Required]
+	public string Note { get; set; } = null!;
+
+	[Required]
+	public Guid ApproverId { get; set; }
+
+	[JsonIgnore]
+	public User? Approver { get; set; }
```
zysim, line 22:
```suggestion
	public User Approver { get; set; } = null!;
```

LeaderboardBackend/Models/Entities/Leaderboard.cs
```diff
@@ -0,0 +1,26 @@
...
+	public string? Rules { get; set; }
+
+	[JsonIgnore] 
+	public List<Ban>? Bans { get; set; }
+
+	[JsonIgnore] 
+	public List<Category>? Categories { get; set; }
+
+	[JsonIgnore] 
+	public List<Modship>? Modships { get; set; }
```
zysim, line 25: Kinda feel like we should initialise these to be empty Lists instead of nullable ones after all. Would be easier to manage. What do you think?

LeaderboardBackend/Models/Entities/Run.cs
```diff
@@ -0,0 +1,28 @@
...
+	[Required]
+	public DateTime Played { get; set; }
+
+	[Required]
+	public DateTime Submitted { get; set; }
+
+	[Required]
+	public RunStatus Status { get; set; }
+
+	public List<Judgement>? Judgements { get; set; }
```
zysim, line 27: Here as well (initialise to an empty List)

LeaderboardBackend/Models/Entities/User.cs
```diff
@@ -0,0 +1,33 @@
...
+	[JsonIgnore] 
+	[InverseProperty("BanningUser")] 
+	public List<Ban>? BansGiven { get; set; }
+
+	[JsonIgnore] 
+	[InverseProperty("BannedUser")] 
+	public List<Ban>? BansReceived { get; set; }
+
+	[JsonIgnore] 
+	public List<Modship>? Modships { get; set; }
```
zysim, line 32: Ditto

LeaderboardBackend/Models/Entities/Judgement.cs
RageCage64: Discussed offline, all `JsonIgnore` properties need to be nullable. That being said, have a completely alternate solution coming in another PR.

LeaderboardBackend/Models/Entities/Leaderboard.cs
RageCage64: Discussed offline but copying the reasoning for not doing this here:

An empty list would be more confusing because empty list and null would be conflated. They mean different things:
* null = we haven't fetched the relationship
* empty list = there are no relationships
Those two things need to be different, so I think nullable list is right

PR #72, opened 2022-03-26
Removed integration project, replaced unit tests

RageCage64:
As I continued to improve Integration tests, I realized that they are
going to be essentially as effective, and (when using an in memory db)
barely slower than unit tests. Might as well save ourselves tons of
effort and have minimal tests that basically just test the API as if it
were a real client.


LeaderboardBackend.Test/TestUsers.cs
```diff
@@ -60,16 +63,63 @@ public static async Task GetUser_Found()
...
+	[Test]
+	public static async Task Me_Unauthorized()
+	{
+		HttpRequestMessage meRequest = new(HttpMethod.Get, "/api/users/me");
+		HttpResponseMessage meResponse = await ApiClient.SendAsync(meRequest);
+		Assert.AreEqual(HttpStatusCode.Forbidden, meResponse.StatusCode);
+	}
+
 	[Test]
 	public static async Task FullAuthFlow()
```
zysim, line 114: Gotta make a point to:
1. Implement logging out in the controller; and
2. Test logging out in this test method as well.

zysim (changes requested):
Nit: for the test files (TestCategories, TestLeaderboards, TestUsers); we could remove the `Test` prefix since it's obvious what these files are.

zysim (approved):
Made one more commit that removes the reference to the integration project in LeaderboardBackend.csproj

PR #74, opened 2022-03-26
Don't overwrite existing env vars

RageCage64:
This will be useful for CI because we can pass stuff into the command
that actually runs tests rather than needing to mess with a .env file in
the CI script.


zysim (approved):
:o

PR #75, opened 2022-03-26
Add GitHub Actions

RageCage64:
I just copy pasted the workflow from the old Go repo and switched it to
use dotnet lol


zysim (dismissed):
lol

Issue #44, opened 2022-02-14
MVP: Judgements


zysim:
🔨🔨🔨

From the [MVP](https://github.com/leaderboardsgg/leaderboard-backend-poc/wiki/MVP-Overview#judgement):

## Model/Relationships

- [x] Create Model - refer to MVP (https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/45)
  - Zero or more Judgements to a Run
  - Zero or more Judgements to a User (must be a moderator)

## Actions

- [x] GET fields with Judgement ID (https://github.com/leaderboardsgg/leaderboard-backend/pull/82)
- [x] POST Judgement with: (https://github.com/leaderboardsgg/leaderboard-backend/pull/82)
  - Run ID;
  - Note (can be empty `''` **iff** Approved == true);
  - Approved (optional; `null` == just a comment)
  - **Condition**: Judgement NOT created if one or more empty Participations exist; i.e. Run has status `Created`
- [x] Judgements **cannot** be edited (https://github.com/leaderboardsgg/leaderboard-backend/pull/82)
- [x] Tests for these (https://github.com/leaderboardsgg/leaderboard-backend/pull/82)

Append each item with the PR that solves it.

PR #82, opened 2022-05-04
Create JudgementsController

zysim:
Closes: #44

Changes:
* `*_approver_id` fields to `*_mod_id` in the DB, along with `ApproverId` and `Approver` to `ModId` and `Mod` in the backend
    * Renaming to this makes more sense IMO, as these can be rejections
* Renamed `Timestamp` to `CreatedAt` (+ migration too)
* Add code to add judgements

### DIFF
--- a/LeaderboardBackend/Controllers/JudgementsController.cs
+++ b/LeaderboardBackend/Controllers/JudgementsController.cs
@@ -3,11 +3,14 @@ using LeaderboardBackend.Controllers.Annotations;
 using LeaderboardBackend.Models.Entities;
 using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Services;
+using LeaderboardBackend.ViewModels;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
 namespace LeaderboardBackend.Controllers;
 
+[Route("api/[controller]")]
+[ApiController]
 public class JudgementsController : ControllerBase
 {
 	private readonly ILogger _logger;
@@ -34,7 +37,7 @@ public class JudgementsController : ControllerBase
 	[ApiConventionMethod(typeof(Conventions),
 						 nameof(Conventions.Get))]
 	[HttpGet("{id}")]
-	public async Task<ActionResult<Judgement>> GetJudgement(long id)
+	public async Task<ActionResult<JudgementViewModel>> GetJudgement(long id)
 	{
 		Judgement? judgement = await _judgementService.GetJudgement(id);
 		if (judgement is null)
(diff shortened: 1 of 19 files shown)


zysim:
Getting empty values for Judgement-related objects in the Open API docs and I can't figure out why. I've even tried defining them as `public record`s, after our request objects, but still no luck:
image
image

They should have details of all properties as well as example values.

LeaderboardBackend/ViewModels/Judgements.cs
```diff
@@ -0,0 +1,56 @@
...
+	public readonly string? Note;
+
+	/// <summary>ID of the mod that made this judgement.</summary>
+	public readonly Guid ModId;
+
+	/// <summary>ID of the related run.</summary>
+	public readonly Guid RunId;
+
+	public JudgementViewModel(
+		long id,
```
RageCage64, line 41: Any reason you don't just take a `Judgement` as an argument here? Seems like a lot of unnecessary work for yourself to take all the args separately

zysim: Actually yeah I don't see why not.

zysim:
I get a weird 404 when try creating a judgement with a valid run and mod user.

LeaderboardBackend/Controllers/JudgementsController.cs
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	/// <summary>Creates a judgement for a run.</summary>
+	/// <response code="201">The created judgement.</response>
+	/// <response code="400">The request body is malformed.</response>
+	/// <response code="403">The run has pending participations.</response>
+	/// <response code="404">For an invalid judgement.</response>
+	[ApiConventionMethod(typeof(Conventions),
+						nameof(Conventions.Post))]
+	[ProducesResponseType(StatusCodes.Status403Forbidden)]
```
RageCage64, line 59: This produces the warning mentioned below. Mixing `ApiConventionMethod` and `ProducesResponseType` is apparently not liked.

RageCage64:
>I get a weird 404 when try creating a judgement with a valid run and mod user.

As discussed offline, I was able to find the fix for this and we decided I should just commit the fix since I got it working.

LeaderboardBackend/Controllers/JudgementsController.cs
zysim: So I had a suspicion, and I was correct.

You just need to have `ProducesResponseType` be above `ApiConventionMethod`. Lmao

RageCage64 (approved):

LeaderboardBackend/Models/Entities/ApplicationContext.cs
```diff
@@ -11,8 +11,15 @@ public ApplicationContext(DbContextOptions<ApplicationContext> options)
...
-	public DbSet<Run> Runs { get; set; } = null!; 
-	public DbSet<Participation> Participations { get; set; } = null!; 
+	public DbSet<Modship> Modships { get; set; } = null!;
+	public DbSet<Run> Runs { get; set; } = null!;
+	public DbSet<Participation> Participations { get; set; } = null!;
 	public DbSet<User> Users { get; set; } = null!;
+
+	protected override void OnModelCreating(ModelBuilder modelBuilder)
+	{
+		modelBuilder.Entity<Judgement>()
```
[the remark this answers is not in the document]
RageCage64: will do this later

Issue #81, opened 2022-05-04
Task: Code Debt: Clean Up Attributes on Controller Actions


zysim:
There's tidying up we need to do for attributes on controller actions:
- BansController.cs: Replace existing Produces*ResponseType attributes for ApiConventionMethod attributes
- RunsController.cs: Remove all ProducesResponseType attributes in all methods, as they are redundant due to the ApiConventionMethod attributes

### Affected Functionality
There shouldn't be any changes. Functionality should stay the same.

### Other Relevant Issues
\-

PR #84, opened 2022-05-14
Issue81 controller attributes

markflm:
Issue 81: https://github.com/leaderboardsgg/leaderboard-backend/issues/81

Removed/replaced any ProduceResponseType attributes in BansController.cs and RunsController.cs with ApiConventionMethod attributes.

added a comment to the readme for something that tripped me up during setup.

added some more detail to a comment pointing to an example of CompareAttribute in the UsersController.

### DIFF
--- a/LeaderboardBackend/Controllers/RunsController.cs
+++ b/LeaderboardBackend/Controllers/RunsController.cs
@@ -24,8 +24,6 @@ namespace LeaderboardBackend.Controllers
 		/// <response code="404">If no Run is found with the provided ID.</response>
 		[ApiConventionMethod(typeof(Conventions),
 							 nameof(Conventions.Get))]
-		[ProducesResponseType(StatusCodes.Status200OK)]
-		[ProducesResponseType(StatusCodes.Status404NotFound)]
 		[HttpGet("{id}")]
 		public async Task<ActionResult<Run>> GetRun(Guid id)
 		{
@@ -41,9 +39,6 @@ namespace LeaderboardBackend.Controllers
 		/// <summary>Creates a Run.</summary>
 		[ApiConventionMethod(typeof(Conventions),
 							 nameof(Conventions.Post))]
-		[ProducesResponseType(StatusCodes.Status201Created)]
-		[ProducesResponseType(StatusCodes.Status400BadRequest)]
-		[ProducesDefaultResponseType]
 		[HttpPost]
 		public async Task<ActionResult> CreateRun([FromBody] CreateRunRequest request)
 		{
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Then run the following commands to run the DB migrations:
 # Install project tools
 dotnet tool restore
 # Run migrations
+# You'll need to cd into the project you want to run the datbase migration for - either LeaderboardBackend or LeaderboardBackend.Test
 dotnet ef database update
 ```
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Then run the following commands to run the DB migrations:
 # Install project tools
 dotnet tool restore
 # Run migrations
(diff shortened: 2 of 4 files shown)


markflm:
I updated the 2 controllers mentioned in the issue but I notice ProduceResponseType is still used in some other ones. should I just remove/replace those with ApiConventionMethod too?

README.md
```diff
@@ -117,6 +117,7 @@ Then run the following commands to run the DB migrations:
 # Install project tools
 dotnet tool restore
 # Run migrations
+# You'll need to cd into the project you want to run the datbase migration for - either LeaderboardBackend or LeaderboardBackend.Test
```
RageCage64, line 120:
```suggestion
cd LeaderboardBackend
```
Migrations are only run in that project anyway, either we say this or we add the project flag to the database update command below.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -1,49 +1,44 @@
...
-		public async Task<ActionResult<Run>> GetRun(Guid id)
-		{
-			Run? run = await _runService.GetRun(id);
-			if (run == null)
-			{
-				return NotFound();
-			}
-
-			return Ok(run);
+		public async Task<ActionResult<Run>> GetRun(Guid id)
```
RageCage64, line 28: Why did these change? I wish GitHub let you check if the change is because of spaces vs tabs.

RageCage64 (commented):
There are a lot of random changes here, I'm wondering if there's a whitespace configuration issue. Does our README mention setting up editorconfig for your editor? It's set up by default for Visual Studio but VSCode and others you'd need to install an extension.

markflm:
> There are a lot of random changes here, I'm wondering if there's a whitespace configuration issue. Does our README mention setting up editorconfig for your editor? It's set up by default for Visual Studio but VSCode and others you'd need to install an extension.

There aren't specific instructions for setting it up with VSCode, but Google helped me get it set up with the C#/OmniSharp extension.

zysim:
> I updated the 2 controllers mentioned in the issue but I notice ProduceResponseType is still used in some other ones. should I just remove/replace those with ApiConventionMethod too?

The real idea is to set a standard where all controller actions use the ApiConventionMethod attributes, only adding ProducesResponseType attributes if really needed.

Man I didn't even realise we actually have so many more actions to clean up; it might be good to break this issue up into smaller chunks.

LeaderboardBackend/Controllers/RunsController.cs
markflm: From what I can tell the file's still using tabs. I can change the values in the .editorconfig and it will take effect on save in VSCode, which makes me think I've got it set up right.

Maybe there's something other than the editorconfig contributing to the formatting. I'll look into it.

RageCage64: If that's the case then it's reasonable to assume GitHub is being pissy I think. I don't actually see them doing anything specifically wrong so if something is weird we can fix it later.

RageCage64:
If it's not too difficult we can do it in this PR, but otherwise I'm fine to do it in more if we want to get this one merged.

RageCage64 (approved):
Going to assume the weird diffs are either GitHub being temperamental or perhaps were mixed spaces and tabs previously that have been corrected.

markflm:
> If it's not too difficult we can do it in this PR, but otherwise I'm fine to do it in more if we want to get this one merged. Either way I'll approve it for now.

Probably best to just get this one merged IMO, don't know if I could get to the rest of them today and I don't want to prevent someone else from taking care of it. 
I think my "workflow" needs to be approved before it can be merged in.

zysim:
GH does have issues with rendering whitespace; your safest bet is to check the code out yourself and look at it. Which I just have. And it's chill.

zysim:
Thanks Kiff! :)

PR #87, opened 2022-05-19
Remove modship

### LINKED ISSUE #36: MVP: Modships
🔨🔨🔨

From the [MVP](https://github.com/leaderboardsgg/leaderboard-backend-poc/wiki/MVP-Overview):

- [x] Create Model - refer to MVP (https://github.com/leaderboardsgg/leaderboard-backend-poc/pull/37)
- [x] A site admin can appoint a user as a moderator to a leaderboard by providing a leaderboard ID and a user ID (#61)
- [ ] A site admin can unappoint a user as a moderator of a leaderboard by providing a leaderboard ID and a user ID
- [ ] Tests for these

Append each item with the PR that solves it.

### Affected Functionality

- Leaderboards: `LeaderboardId` foreign key links to a Leaderboard's ID
- Users: `UsersId` foreign key links to a Leaderboard's ID

### Other Relevant Issues
(To be filled)

ayecptn:
Closes #36 

Changes:
- Added endpoint to remove a modship
- Added Tests
- Refactored old Modship-Test


LeaderboardBackend.Test/Modships.cs
```diff
@@ -35,41 +36,111 @@ public static async Task MakeMod_Success()
...
+		// Make user a mod
+		Modship created = await CreateModship(createdLeaderboard.Id, jwt);
+
+		// Confirm that user is a mod
+		Modship retrieved = await GetModship(jwt);
+
+		// Remove the modship
+		await RemoveModship(createdLeaderboard.Id, jwt);
+
+		try
```
RageCage64, line 64: I think instead of returning from the catch block and throwing an exception after, it would be better to `Assert.Fail` as the second line of the try block. With `Assert.Fail` you can actually give it a nicer message to describe the expected behaviour too.

LeaderboardBackend/Services/Impl/ModshipService.cs
```diff
@@ -30,8 +30,14 @@ public async Task<List<Modship>> LoadUserModships(Guid userId)
 					.ToListAsync();
 	}
 
-	// TODO: Implement this
-	// public async Task DeleteModship(Modship modship)
-	// {
-	// }
+	public async Task DeleteModship(Modship modship)
+	{
+		var deleteModship = _applicationContext.Modships.Where(m =>
```
RageCage64, line 35: Not your fault because we forgot to write down in our code style but we're opting not to use `var`. Would you be able to change this to a `Modship?`?

LeaderboardBackend/Services/Impl/ModshipService.cs
```diff
@@ -30,8 +30,14 @@ public async Task<List<Modship>> LoadUserModships(Guid userId)
...
+			m.LeaderboardId == modship.LeaderboardId &&
+			m.UserId == modship.UserId)
+			.FirstOrDefault() ?? throw new NullReferenceException();
```
RageCage64, line 38: I don't know if `NullReferenceException` is the way to go here, cause I don't think it accurately describes the issue. The closest exception I can think of in the `System` namespace is `KeyNotFoundException` but maybe a custom exception would be a good idea for this? Trying to think critically because in all likelihood, all our future Delete operations will be based on whatever you do here.

RageCage64 (commented):
Thank you for the PR! Looks great overall. Made some small suggestions.

LeaderboardBackend/Controllers/ModshipsController.cs
```diff
@@ -76,4 +76,43 @@ public async Task<ActionResult> MakeMod([FromBody] CreateModshipRequest body)
...
+	{
+		User? user = await _userService.GetUserById(body.UserId);
+		Leaderboard? leaderboard = await _leaderboardService.GetLeaderboard(body.LeaderboardId);
+
+		if (user is null || leaderboard is null)
+		{
+			return NotFound();
+		}
+
+		Modship modship = new()
```
RageCage64, line 99: @zysim had a good suggestion for this: we should add something to the `ModshipService` that looks up a Modship by UserId and LeaderboardId, then use that here. It makes a lot more sense, because we can check for null here. If this passes a null check, then all the delete has to do is mark it, rather than needing to worry about doing a lookup and throwing an exception.

So if possible:
1. Add a method to the ModshipService like `GetModshipForLeaderboard`
2. Have that method return null if nothing is found
3. Do a null check here and `return NotFound` if it is (similar to other lookup actions)
4. Make `DeleteModship` just do the marking deleted part

I think that would put us in a pretty good spot for this, and would be a good model for future deletes.

RageCage64 (changes requested):

LeaderboardBackend/Services/Impl/ModshipService.cs
RageCage64: Probably obsoleted by the new requested change

LeaderboardBackend/Services/Impl/ModshipService.cs
RageCage64: Probably obsoleted by new Requested Change

LeaderboardBackend/Controllers/ModshipsController.cs
ayecptn: Good points. Changed as suggested.

RageCage64 (approved):

PR #88, opened 2022-05-21
Fixed all inconsistencies to match style guide

RageCage64:
Have fun Shep

### DIFF
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -3,10 +3,8 @@ using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Test.Lib;
 using LeaderboardBackend.Test.TestApi;
 using LeaderboardBackend.Test.TestApi.Extensions;
-using Microsoft.AspNetCore.Authentication.JwtBearer;
 using NUnit.Framework;
 using System.Net;
-using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace LeaderboardBackend.Test;
@@ -18,7 +16,7 @@ internal class Categories
 	private static TestApiClient ApiClient = null!;
 	private static string? Jwt;
 
-	[SetUp]
+	[OneTimeSetUp]
 	public static async Task SetUp()
 	{
 		Factory = new TestApiFactory();
@@ -44,7 +42,7 @@ internal class Categories
 	{
 		RequestFailureException e = Assert.ThrowsAsync<RequestFailureException>(async () =>
 			await ApiClient.Get<Category>(
-				$"/api/categories/1",
+				$"/api/categories/69",
 				new()
 				{
 					Jwt = Jwt,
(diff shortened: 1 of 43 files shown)


LeaderboardBackend.Test/Categories.cs
```diff
@@ -44,7 +42,7 @@ public static void GetCategory_NotFound()
 	{
 		RequestFailureException e = Assert.ThrowsAsync<RequestFailureException>(async () =>
 			await ApiClient.Get<Category>(
-				$"/api/categories/1",
+				$"/api/categories/69",
 				new()
```
zysim, line 46: I cannot believe you

LeaderboardBackend/Controllers/Annotations/Conventions.cs
```diff
@@ -6,6 +6,7 @@ namespace LeaderboardBackend.Controllers.Annotations;
 public static class Conventions
 {
 	[ProducesResponseType(StatusCodes.Status200OK)]
+	[ProducesResponseType(StatusCodes.Status400BadRequest)]
 	[ProducesResponseType(StatusCodes.Status404NotFound)]
```
zysim, line 10: Adding an attribute here will mean that ASP.NET expects doc comments for every action that uses this convention. We sure we want this?

LeaderboardBackend/Controllers/Annotations/Conventions.cs
```diff
@@ -20,6 +21,13 @@ public static class Conventions
 	public static void Post(params object[] parameters)
 	{ }
 
+	[ProducesResponseType(StatusCodes.Status200OK)]
+	[ProducesResponseType(StatusCodes.Status400BadRequest)]
+	[ProducesResponseType(StatusCodes.Status404NotFound)]
+	[ProducesResponseType(StatusCodes.Status403Forbidden)]
```
zysim, line 27: Ditto

LeaderboardBackend/Models/Requests/JudgementRequests.cs
```diff
@@ -3,12 +3,24 @@
...
-/// <param name="Approved">
-///   The judgement result. Can be true, false, or null. For the latter two, <paramref name="Note"/> must be non-empty.
-/// </param>
-public readonly record struct CreateJudgementRequest(Guid RunId, [Note] string Note, bool? Approved);
+public record CreateJudgementRequest
+{
+
+	/// <summary>GUID of the run.</summary>
+	/// <example>e1c010d7-b499-4196-be17-aa27a053f3dc</example>
+	public Guid RunId { get; set; }
```
zysim, line 11: No need to change this, but GUIDs get generated in the docs. There's no need to set an explicit example.

zysim (changes requested):

LeaderboardBackend/Controllers/Annotations/Conventions.cs
RageCage64: Guess we could make a new convention instead?

zysim: I'm not sure if it's worth creating a new convention. Guess we could add one for now? Could always remove it if needed

RageCage64: Decided to leave this, because theoretically all routes can produce 400 and the docs get generated anyway

zysim (approved):
Lovely. Absolutely lovely.

We might want to review our Style Guide just in case there are conflicts.

PR #89, opened 2022-05-22
Ran dotnet format instead of just style

RageCage64:
I had only run dotnet format style and not dotnet format itself, which
caught more things. Also fixed a thing on JudgementViewModel to get rid
of a nullable warning.


LeaderboardBackend/Models/ViewModels/JudgementViewModel.cs
```diff
@@ -17,7 +17,7 @@ public record JudgementViewModel
 	///   When the judgement was made. Follows <a href="https://www.ietf.org/rfc/rfc3339.txt">RFC 3339</a>.
 	/// </summary>
 	/// <example>2022-01-01T12:34:56Z / 2022-01-01T12:34:56+01:00</example>
-	public string CreatedAt { get; set; }
+	public string? CreatedAt { get; set; }
```
zysim, line 20: A little iffy about this one since `CreatedAt` shouldn't ever be `null`

zysim (commented):
Just thought of something extra;

On top of reviewing our style guide, we should add `dotnet format` to our CI. *Possibly* with `--exclude Migrations`.

zysim:
> Just thought of something extra;
> 
> On top of reviewing our style guide, we should add `dotnet format` to our CI. _Possibly_ with `--exclude Migrations`.

With the former, frankly we don't have to do it like *now* now, since there *are* lots of rules in em. Could always update the style guide if a future conflict pops up. And I can add the build step tomorrow if you don't already do it.

LeaderboardBackend/Models/ViewModels/JudgementViewModel.cs
RageCage64: It doesn't really matter, cause this is an output model. We don't do any data validation on it.

RageCage64:
The migrations being formatted does not appear to have any adverse affects, considering the tests all still work as expected.

LeaderboardBackend/Models/ViewModels/JudgementViewModel.cs
zysim: Mmmmmmmmmm fine

zysim:
Oh yeah I'm sure everything works. In fact I expect it to too. I'm just probably thinking too far ahead where it might slow down builds when we have a lot of migration files.

zysim (approved):

PR #90, opened 2022-05-23
Created Jobs framework

RageCage64:
Added new framework for creating and running backend jobs. These will be
used by developers to quickly scaffold data for their environments, or
to do important tasks on a running server.

Also fixed a weird data validation thing I discovered while creating and
testing this feature.

### DIFF
--- a/LeaderboardBackend.Jobs/Core/Options.cs
+++ b/LeaderboardBackend.Jobs/Core/Options.cs
@@ -1,0 +1,77 @@
+using System.ComponentModel.DataAnnotations;
+using LeaderboardBackend.Models.Entities;
+using LeaderboardBackend.Models.Requests;
+using BCryptNet = BCrypt.Net.BCrypt;
+
+namespace LeaderboardBackend.Jobs.Core;
+
+internal static class Options
+{
+	public static string StringLine(string optionName, Func<string?, bool>? validator = null, string invalidMessage = "Invalid input.")
+	{
+		if (validator is null)
+		{
+			validator = s => !string.IsNullOrWhiteSpace(s);
+		}
+
+		string? input = null;
+		do
+		{
+			Console.Write($"{optionName}: ");
+			input = Console.ReadLine();
+			if (!validator(input))
+			{
+				Console.WriteLine(invalidMessage);
+			}
+		} while (!validator(input));
+
+		return input ?? "";
+	}
+
+	public static bool YesOrNo(string question)
+	{
+		Console.Write($"{question} (y to say yes): ");
+		string? input = Console.ReadLine();
+		return input is not null && input.Trim() == "y";
+	}
+
+	public static User User()
+	{
+		Console.WriteLine("Please provide the new user's information.");
+
+		RegisterRequest request = new();
+		bool valid = false;
+		do
+		{
+			string username = StringLine("Username");
+			string email = StringLine("Email");
+			string password = StringLine("Password");
+			string passwordConfirm = StringLine(
+				"Confirm Password",
+				validator: s => !string.IsNullOrWhiteSpace(s) && s == password,
+				invalidMessage: "Passwords must match"
+			);
+			request = new()
+			{
+				Username = username,
+				Email = email,
+				Password = password,
+				PasswordConfirm = passwordConfirm,
+			};
+			List<ValidationResult> errors = new();
+			ValidationContext ctx = new(request);
+			valid = Validator.TryValidateObject(request, ctx, errors, validateAllProperties: true);
+			if (errors.Count > 0)
+			{
+				Console.WriteLine($"Entity was invalid: {string.Join('/', errors.Select(r => r.ErrorMessage))}");
+			}
+		} while (!valid);
+
+		return new User
+		{
+			Username = request.Username,
+			Email = request.Email,
+			Password = BCryptNet.EnhancedHashPassword(request.Password)
+		};
+	}
+}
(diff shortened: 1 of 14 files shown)


RageCage64:
Screenshot for running the `user` job as an example. (Unrelated: demonstrates that the User validation regex is probably zonked)
image

zysim:
This is so cool

LeaderboardBackend.Jobs/Jobs/ScaffoldLeaderboardAndMod.cs
```diff
@@ -0,0 +1,55 @@
...
+		{
+			Name = name,
+			Slug = slug,
+		};
+		Context.Leaderboards.Add(newLeaderboard);
+		await Context.SaveChangesAsync();
+		long? newLeaderboardId = Context.Leaderboards.Where(l => l.Name == name).FirstOrDefault()?.Id;
+		if (newLeaderboardId is null)
+		{
+			throw new Exception("Unexpected error creating leaderboard.");
```
zysim, line 35: You know if - assuming this happens - that the DB row would still exist? I'm wondering if we'd need to worry about having to clean invalid rows if this ever happens. Same goes for L44, for creating users.

In fact; _would_ this ever even happen? Cuz if so, we should also add this check in `ScaffoldUser.Run()` too

LeaderboardBackend.Jobs/Jobs/ScaffoldUser.cs
```diff
@@ -0,0 +1,28 @@
...
+	private ApplicationContext Context;
+
+	public ScaffoldUser(ApplicationContext context)
+	{
+		Context = context;
+	}
+
+	public string CommandName => "user";
+	public string Description => "Create a new user.";
+	public override string ToString() => $"{CommandName}: {Description}";
```
zysim, line 19: Could we add these fields to the interface instead? Or would (for lack of a better term) adding `ToString` mean that `IJob` would have to be an abstract class? Combination request and query about basic C# here

LeaderboardBackend.Jobs/Core/Options.cs
```diff
@@ -0,0 +1,77 @@
...
+				Console.WriteLine(invalidMessage);
+			}
+		} while (!validator(input));
+
+		return input ?? "";
+	}
+
+	public static bool YesOrNo(string question)
+	{
+		Console.Write($"{question} (y to say yes): ");
```
zysim, line 33: We could do the conventional `y/[N]` thing instead of `y to say yes`

LeaderboardBackend.Jobs/Jobs/ScaffoldUser.cs
```diff
@@ -0,0 +1,28 @@
...
+
+	async Task IJob.Run()
+	{
+		User newUser = Options.User();
+		newUser.Admin = Options.YesOrNo("Will this user be an admin?");
+		Context.Users.Add(newUser);
+		await Context.SaveChangesAsync();
+	}
```
zysim, line 27: Refer to comment in `ScaffoldLeaderboardAndMod.cs`

LeaderboardBackend/Models/Requests/CategoryRequests.cs
```diff
@@ -4,7 +4,7 @@
 namespace LeaderboardBackend.Models.Requests;
 
 /// <summary>Request object sent when creating a Category.</summary>
-public record CreateCategoryRequest
+public class CreateCategoryRequest
```
zysim, line 7: Why was this changed to a class?

zysim (changes requested):

LeaderboardBackend.Jobs/Jobs/ScaffoldUser.cs
RageCage64: These fields are on the interface

LeaderboardBackend.Jobs/Jobs/ScaffoldLeaderboardAndMod.cs
RageCage64: This wouldn't happen I'm just appeasing null reference type warnings

LeaderboardBackend.Jobs/Core/Options.cs
RageCage64: Makes the code more annoying to check explicitly for `y/n` and I figured it wasn't a big enough deal

LeaderboardBackend/Models/Requests/CategoryRequests.cs
RageCage64: accident

LeaderboardBackend.Jobs/Jobs/ScaffoldUser.cs
RageCage64: Changed the Run() to implement implicitly so they are all consistent

LeaderboardBackend.Jobs/Jobs/ScaffoldLeaderboardAndMod.cs
zysim: Mmm

LeaderboardBackend.Jobs/Core/Options.cs
zysim: Oh nah cuz `[N]` implies replying with nothing means a rejection, and since we only check for ys, it works out. It's a nitpick, of course

zysim (dismissed):
Depends if you wanna resolve my last remaining comment. I'm happy to get this merged in.

LeaderboardBackend.Jobs/Core/Options.cs
RageCage64: True, I didn't think of it that way. I'll do that.

PR #91, opened 2022-05-23
Allow Numbers in Usernames

zysim:
Allows numbers to be used in names

### DIFF
--- a/LeaderboardBackend/Models/Entities/User.cs
+++ b/LeaderboardBackend/Models/Entities/User.cs
@@ -14,7 +14,7 @@ public class User
 	/// The User's name. Must be:
 	/// <ul>
 	///   <li>between 2 - 25 characters inclusive; and</li>
-	///   <li>a sequence of letters, each separated by zero or one of:</li>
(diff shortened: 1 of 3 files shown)


LeaderboardBackend/Models/Entities/User.cs
```diff
@@ -14,7 +14,7 @@ public class User
 	/// The User's name. Must be:
 	/// <ul>
 	///   <li>between 2 - 25 characters inclusive; and</li>
-	///   <li>a sequence of letters, each separated by zero or one of:</li>
+	///   <li>a sequence of letters or numbers, each separated by zero or one of:</li>
```
RageCage64, line 17: nit: "An alphanumeric sequence" maybe? I don't think anyone's really gonna be that confused by this wording though

RageCage64 (approved):

LeaderboardBackend/Models/Entities/User.cs
zysim:
```suggestion
	///   <li>an alphanumeric sequence, each separated by zero or one of:</li>
```

PR #98, opened 2022-06-04
Added AuthService method to get jwt claims

RageCage64:
Added `GetUserIdFromClaims` and `GetEmailFromClaims` to the AuthService,
and use those instead of a UserService method to get the user from
claims. This makes for a more sane pattern throughout controllers using
claim data, and reduces database calls.

### DIFF
--- a/LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
+++ b/LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
@@ -12,17 +12,20 @@ public class UserTypeAuthorizationHandler : AuthorizationHandler<UserTypeRequire
 	private readonly TokenValidationParameters JwtValidationParams;
 	private readonly IUserService UserService;
 	private readonly IModshipService ModshipService;
+	private readonly IAuthService AuthService;
 
 	public UserTypeAuthorizationHandler(
 		IConfiguration config,
 		IModshipService modshipService,
-		IUserService userService
+		IUserService userService,
+		IAuthService authService
 	)
 	{
 		JwtHandler = JwtSecurityTokenHandlerSingleton.Instance;
 		JwtValidationParams = TokenValidationParametersSingleton.Instance(config);
 		UserService = userService;
 		ModshipService = modshipService;
+		AuthService = authService;
 	}
 
 	protected override Task HandleRequirementAsync(
@@ -35,7 +38,13 @@ public class UserTypeAuthorizationHandler : AuthorizationHandler<UserTypeRequire
 			return Task.CompletedTask;
 		}
 
-		User? user = UserService.GetUserFromClaims(context.User).Result;
+		Guid? userId = AuthService.GetUserIdFromClaims(context.User);
+		if (userId is null)
+		{
+			context.Fail();
+			return Task.CompletedTask;
+		}
+		User? user = UserService.GetUserById(userId.Value).Result;
 
 		if (user is null || !Handle(user, context, requirement))
 		{
--- a/LeaderboardBackend/Controllers/JudgementsController.cs
+++ b/LeaderboardBackend/Controllers/JudgementsController.cs
@@ -18,18 +18,21 @@ public class JudgementsController : ControllerBase
 	private readonly IJudgementService JudgementService;
 	private readonly IRunService RunService;
 	private readonly IUserService UserService;
+	private readonly IAuthService AuthService;
 
 	public JudgementsController(
 		ILogger<JudgementsController> logger,
 		IJudgementService judgementService,
 		IRunService runService,
-		IUserService userService
+		IUserService userService,
+		IAuthService authService
 	)
 	{
 		Logger = logger;
 		JudgementService = judgementService;
 		RunService = runService;
 		UserService = userService;
+		AuthService = authService;
 	}
 
 	/// <summary>Gets a Judgement from its ID.</summary>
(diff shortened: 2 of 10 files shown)


LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
```diff
@@ -35,7 +38,13 @@ UserTypeRequirement requirement
...
 		}
 
-		User? user = UserService.GetUserFromClaims(context.User).Result;
+		Guid? userId = AuthService.GetUserIdFromClaims(context.User);
+		if (userId is null)
+		{
+			context.Fail();
+			return Task.CompletedTask;
+		}
+		User? user = UserService.GetUserById(userId.Value).Result;
```
zysim, line 47: Do we need `.Value` here? Can't we just use `userId`?

LeaderboardBackend/Controllers/JudgementsController.cs
```diff
@@ -60,7 +63,11 @@ public async Task<ActionResult<JudgementViewModel>> GetJudgement(long id)
 	[HttpPost]
 	public async Task<ActionResult<JudgementViewModel>> CreateJudgement([FromBody] CreateJudgementRequest body)
 	{
-		User? mod = await UserService.GetUserFromClaims(HttpContext.User);
+		Guid? modId = AuthService.GetUserIdFromClaims(HttpContext.User);
+		if (modId is null)
+		{
+			return Forbid();
```
zysim, line 69: Problem is we already have user validation done in our middleware, so this bit shouldn't be necessary.

I believe we're running into the same problem here just in a different form. Since we'd already catch and handle the null case in authZ already.

In fact I almost want to suggest we 500 here instead, since this shouldn't ever happen.

LeaderboardBackend/Controllers/ParticipationsController.cs
```diff
@@ -81,12 +84,12 @@ public async Task<ActionResult> CreateParticipation([FromBody] CreateParticipati
 	[HttpPut]
 	public async Task<ActionResult> UpdateParticipation([FromBody] UpdateParticipationRequest request)
 	{
-		User? user = await UserService.GetUserFromClaims(HttpContext.User);
-		if (user is null)
+		Guid? userId = AuthService.GetUserIdFromClaims(HttpContext.User);
+		if (userId is null)
 		{
 			return Forbid();
```
zysim, line 90: Ditto

LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -120,7 +120,16 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest bod
 	[HttpGet("me")]
 	public async Task<ActionResult<User>> Me()
 	{
-		User? user = await UserService.GetUserFromClaims(HttpContext.User);
-		return user is not null ? Ok(user) : Forbid();
+		string? email = AuthService.GetEmailFromClaims(HttpContext.User);
+		if (email is null)
+		{
+			return Forbid();
```
zysim, line 126: Ditto

LeaderboardBackend/Services/IAuthService.cs
```diff
@@ -1,8 +1,11 @@
+using System.Security.Claims;
 using LeaderboardBackend.Models.Entities;
 
 namespace LeaderboardBackend.Services;
 
 public interface IAuthService
 {
 	string GenerateJSONWebToken(User user);
+	public string? GetEmailFromClaims(ClaimsPrincipal claims);
+	public Guid? GetUserIdFromClaims(ClaimsPrincipal claims);
```
zysim: I feel that we should rename this to `GetIdFromClaims`; should be obvious we're getting the user's ID, right?

LeaderboardBackend/Services/Impl/AuthService.cs
```diff
@@ -37,4 +37,26 @@ public string GenerateJSONWebToken(User user)
...
+		string? userIdStr = claims.FindFirstValue(JwtRegisteredClaimNames.Sub);
+		if (userIdStr is null)
+		{
+			return null;
+		}
+
+		Guid? userId = null;
+		try
+		{
+			userId = Guid.Parse(userIdStr);
```
zysim, line 57: What value will `userId` take on `FormatException`?

zysim (changes requested):

LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
RageCage64: `Guid?` is not the same as a nullable reference type. `Guid` is considered a primitive. For those, you have to get `.Value` out of it specifically.

LeaderboardBackend/Services/IAuthService.cs
RageCage64: The claims can contain other IDs in the future. For now it's just the UserId, but making this explicit future proofs it.

LeaderboardBackend/Services/Impl/AuthService.cs
RageCage64: `null`

LeaderboardBackend/Controllers/JudgementsController.cs
RageCage64: >This shouldn't be necessary
There's no way around it with nullable reference types. The compiler can't figure out that this isn't necessary.

>Suggest we 500 here
We could throw some kind of exception, not sure what.

zysim: After thinking about it we can keep things this way. No changes needed here

LeaderboardBackend/Services/IAuthService.cs
zysim: Sure, let's keep it then

zysim (approved):
Got nothing else to suggest. I'll approve n merge

Issue #97, opened 2022-06-04
Task: Update User Type Policy Authorisation Result


zysim:
TL;DR: don't just return 404s for everything.

Currently in MiddlewareResultHandler.cs, we simply return 404s for any fail case. We should revise this to pass the correct authz result through to the controllers.

### Affected Functionality

- The error response codes returned by controllers will be changed.
- This means documentation on controller actions + ProducesResponseType attributes will need to be changed as well.

PR #99, opened 2022-06-06
Remove Custom Authz Result Handler

zysim:
Closes: #97


zysim:
Closing this for now. Realised there's much more to consider than simply deleting the result handler.

PR #103, opened 2022-06-10
Add swagger documentation to `RunStatus` enums.

erunks:


RageCage64 (approved):
Interesting, I didn't realize these came in the API response as a number but it makes sense.

PR #105, opened 2022-06-11
Refactor AuthZ Handler

zysim:
Changes `UserTypeAuthorizationHandler.cs` to add the user role to the ClaimsPrincipal (controller context object) before evaluating authorisation. This means we can simply call `HttpContext.User.HasClaim("role", UserTypes.<whatever>)` in controller actions to get the already-authorised user's role.


zysim:
Setting this as a draft because it's based off of #100.

zysim:
Closing. Contains work affecting `UserTypes`, which should be deprecated in favour of using `UserRole`. However, the work is still relevant. Refer to #170 for the new task, and maybe use this PR as reference.

PR #119, opened 2022-07-14
MVP: Time Metrics

zysim:


zysim:
Pausing on this PR as of this writing. We can't decide how we want to go forward with implementing this model, particularly with the `Min` and `Max` values, if we event want to implement them at all.

PR #121, opened 2022-07-30
Fix up `.editorconfig` and enforce (#117)

### LINKED ISSUE #117: Task: Cleaning up and standardizing code style
The project currently sets some very strange, unexplained, and non-standard rules for its code. These rules are very often not followed by large parts of the project. This issue is to address some of these rules to perhaps change them or at least enforce them across the code.

For reference: the current [style guide](https://github.com/leaderboardsgg/leaderboard-backend/wiki/Style-Guide).

---

It makes little sense to stray from the suggested C# standard on this, which is the following;
```cs
private int _foo;
protected int _bar;
private protected int _baz;

internal int Qux { get; set; }
protected internal int Corge { get; set; }
public int Grault { get; set; }

private static int s_Garply;
public static int Thud { get; set; }
```
`private` or `protected` objects are fields and begin with an underscore, followed by a `camelCase` name. Fields with the `static` modifier also gain a leading `s`.  
`internal` or `public` objects are properties and use `PascalCase`.  
These conventions ensure an easy reading experience, allow things to be understood much quicker and intuitively, and also make Intellisense a bit quicker to pick up on which variable one is looking for. The current `.editorconfig` rules also suggest local and method parameters to use `PascalCase`, which is followed absolutely nowhere in the solution.

Documentation comments should be on every type and method and should explain every parameter in short detail. Currently, documentation comments are lackluster in many parts and are very inconsistently styled. I propose the following styling for all documentation comments;
```cs
/// <summary>
///     A summary for the method.
/// </summary>
/// <remarks>
///     Additional remarks about the method.
/// </remarks>
/// <param name="bar">A parameter for the method.</param>
/// <returns>
///     A description about the return value.
/// </returns>
bool Foo(int bar) { }
```
The above leaves easy room for multiple lines on each of the `<summary/>`, `<remarks/>`, and `<returns/>` tags. The 4-space indentations ensure an easy reading experience and do not affect the displayed values. Additionally, encourage the use of `<see/>` with the `cref`, `href`, and `langword` attributes, as well as the rare use of the `<c/>` and `<i/>` tags. Prefer `<br/>` for line breaks over `<para/>`. For properties that may not be perfectly intuitive, use `<value/>` to elaborate on what their value represents.

It might be beneficial to actually enforce a maximum line length of ~100 columns. Be wary that this may conflict with `<param/>` documentation comments with the syling above, since it does not allow for an easy line break. Its styling may need changing to that of the other tags illustrated above which do allow line breaks.

### Requests & ViewModels
Using `public record` classes is very beneficial for equality checks, and I'm glad we're embracing the concept. However, the current implementations stray quite a bit from the norm. A `record` should be declared like the following;
```cs
record Foo(int Bar, string Qux);
```
The style guide and the code currently declare all properties explicitly, which is simply unnecessary. Parameters specified in the primary constructor of the `record` are required by default (just like with any method's parameters). Additional constructor overloads can be added within the `record`. Using the above styling ensures with an absolute guarantee that all required parameters are passed. No mistakes can be made here.

### Indentations, Method Bodies, Line Breaks
Finally, I would like to speak on the choice of indentations in the project. Tab-stops are a huge hassle to work with and around. This is very ironically shown within the style guide itself under [Controller Actions](https://github.com/leaderboardsgg/leaderboard-backend/wiki/Style-Guide#controller-actions);
```cs
[ApiConventionMethod(typeof(Conventions),
						nameof(Conventions.Get))]
```
I propose changing to the C#-suggested indentations of 4 spaces. This will fix issues such as this and will avoid any discrepancies between what a user with 4-space-tab stops (Visual Studio) and a user with 8-space-tab stops (GitHub) may see.

I have also noticed in parts of the code that occasionally, expression bodies are used for methods. This makes reading such methods very unintuitive and difficult, especially if most of the rest of the solution does not use expression bodies. I suggest that block bodies are enforced throughout the project.

Finally, I want to address some of the very strange and inconsistent line breaks for call chains and method signatures. It may be beneficial to enforce consistent wrapping for long method signatures such as
https://github.com/leaderboardsgg/leaderboard-backend/blob/cbb6b2be50856dd79219f7af1c1e5b759df6ad6a/LeaderboardBackend/Authorization/MiddlewareResultHandler.cs#L12-L17 and
https://github.com/leaderboardsgg/leaderboard-backend/blob/cbb6b2be50856dd79219f7af1c1e5b759df6ad6a/LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs#L17-L23 where the closing parenthesis should be on the same line across the two signatures (in my opinion on the line of the final parameter).  
Call chains, especially ones that use `System.Linq`, should be wrapped and aligned in a sensible way. The following code
https://github.com/leaderboardsgg/leaderboard-backend/blob/cbb6b2be50856dd79219f7af1c1e5b759df6ad6a/LeaderboardBackend/Services/Impl/ParticipationService.cs#L39 should be formatted as
```cs
return await ApplicationContext.Participations
             .Where(p => p.RunId == run.Id)
             .ToListAsync();
```
(Also for readability's sake, the lambda parameter `p` should be named `participation`.)  

Other smaller cases include code like the following;
```cs
if
{

} else
{

}

try
{

} catch
{

}
```
In this example, both `else` and `catch` should be on their own line.

---

I hope this is not too overwhelming of an issue, and I'm happy to work on all of this myself, just let me know what you think.

just-ero:
Addresses most of #117.

`.editorconfig` changes were discussed in the community Discord. This PR aims to enforce the new rules throughout the project.

Major changes include
* new naming conventions that are closer to the C# standard,
* reordering of attributes and fields alphabetically,
* indenting `<summary/>` tags,
* add empty lines where appropriate to break blocks up into sensible segments,
* fix wrapping and line lengths.


zysim (approved):
Yeah, LGTM. Thanks, man

PR #128, opened 2022-10-05
Add BaseEntity Class

zysim:
* Adds `created_at`, `updated_at`, and `deleted_at` fields to all models
* Updates `Equals` methods to match all above timestamp fields
* Updates DeleteModship to soft-delete, along with its test


LeaderboardBackend/Models/Entities/BaseEntity.cs
```diff
@@ -0,0 +1,38 @@
+using System.ComponentModel.DataAnnotations;
+using NodaTime;
+
+/// <summary>
+///     Base class that provides common fields to all Entities.
+/// </summary>
+public class BaseEntity
```
ayecptn, line 7: As mentioned in #129 make this class abstract

LeaderboardBackend/Models/Entities/BaseEntity.cs
```diff
@@ -0,0 +1,38 @@
...
+	/// </summary>
+	[Required]
+	public Instant UpdatedAt { get; set; } = SystemClock.Instance.GetCurrentInstant();
+
+	/// <summary>
+	///     The time this entity was deleted.
+	/// </summary>
+	public Instant? DeletedAt { get; set; }
+
+	public override bool Equals(object? obj)
```
ayecptn, line 26: Make the Equals and GetHashCode also abstract and implemented them in each Entity

ayecptn (changes requested):

LeaderboardBackend/Models/Entities/BaseEntity.cs
zysim: Actually, I just need to remove these lines in `BaseEntity`, as they already are overrides here. The methods are part of the inner `object` type. I'll still do the latter part, of course.

zysim:
Stale PR. Closing.

PR #129, opened 2022-10-09
WIP: Add NumericalMetrics

zysim:
This'll be a WIP for a while. Also it's based on #128


LeaderboardBackend.Test/NumericalMetrics.cs
```diff
@@ -0,0 +1,135 @@
...
+using NUnit.Framework;
+
+namespace LeaderboardBackend.Test;
+
+[TestFixture]
+internal class NumericalMetrics
+{
+	private static TestApiClient s_ApiClient = null!;
+	private static TestApiFactory s_Factory = null!;
+	private static string s_Jwt = null!;
```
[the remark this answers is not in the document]
zysim: I _gotchu_

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -30,14 +30,15 @@ public async Task<ActionResult<Category>> GetCategory(long id)
...
 
-		Category? category = await _categoryService.GetCategory(id);
-
-		if (category == null)
+		try
+		{
+			Category category = await _categoryService.Get(id);
+			return Ok(category);
+		}
+		catch (System.Exception)
```
TheTedder, line 38: Would it be better not to use exceptions for control flow, and instead only use them when something actually goes wrong? The category requested not being found is just something that can happen here, not really an error.

LeaderboardBackend/Models/Entities/BaseEntity.cs
```diff
@@ -0,0 +1,38 @@
...
+
+	public override bool Equals(object? obj)
+	{
+		return obj is BaseEntity entity
+			&& CreatedAt == entity.CreatedAt
+			&& UpdatedAt == entity.UpdatedAt
+			&& DeletedAt == entity.DeletedAt;
+	}
+
+	public override int GetHashCode()
```
TheTedder, line 34: This doesn't seem right at all. Shouldn't the hash code be based on ID?

LeaderboardBackend/Models/Entities/BaseEntity.cs
```diff
@@ -0,0 +1,38 @@
...
+	/// </summary>
+	[Required]
+	public Instant UpdatedAt { get; set; } = SystemClock.Instance.GetCurrentInstant();
+
+	/// <summary>
+	///     The time this entity was deleted.
+	/// </summary>
+	public Instant? DeletedAt { get; set; }
+
+	public override bool Equals(object? obj)
```
TheTedder, line 26: This is wrong isn't it? Shouldn't equality be ID-based?

LeaderboardBackend/Models/Entities/BaseEntity.cs
```diff
@@ -0,0 +1,38 @@
+using System.ComponentModel.DataAnnotations;
+using NodaTime;
+
+/// <summary>
+///     Base class that provides common fields to all Entities.
+/// </summary>
+public class BaseEntity
```
TheTedder, line 7: This class should be abstract.

LeaderboardBackend/Models/Entities/NumericalMetric.cs
```diff
@@ -0,0 +1,96 @@
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics.CodeAnalysis;
+
+namespace LeaderboardBackend.Models.Entities;
+
+/// <summary>
+///     Represents a `NumericalMetric` for a `Category`.
+/// </summary>
+public class NumericalMetric : BaseEntity
```
TheTedder, line 9: These aren't so much numerical metrics as they are integer metrics.

LeaderboardBackend/Models/Entities/NumericalMetric.cs
```diff
@@ -0,0 +1,96 @@
...
+	/// <summary>
+	///     A collection of `Category`s scoped to the `NumericalMetric`.
+	/// </summary>
+	[MinLength(1)]
+	public List<Category>? Categories { get; set; }
+
+	/// <summary>
+	///     A collection of `Run`s scoped to the `NumericalMetric`.
+	/// </summary>
+	public List<Run>? Runs { get; set; }
```
TheTedder, line 46: This doesn't seem right. How can a metric have runs that belong to it? Is this just the runs that are using this metric?

LeaderboardBackend/Models/Entities/Run.cs
```diff
@@ -61,13 +61,18 @@ public class Run
...
 	/// <summary>
-	/// 	Relationship model for `CategoryId`.
+	/// 	The `Category` object itself, which is tied to the `Run`.
 	/// </summary>
 	public Category? Category { get; set; }
+
+	/// <summary>
+	///     A collection of `NumericalMetric`s scoped to the `Run`.
+	/// </summary>
+	public List<NumericalMetric>? NumericalMetrics { get; set; }
```
TheTedder, line 77: This seems wrong. Doesn't a run need to be associated with a numerical metric through a value and not the metric itself?

LeaderboardBackend/Models/Entities/Timestamps.cs
```diff
@@ -0,0 +1,25 @@
+using System.ComponentModel.DataAnnotations;
+using NodaTime;
+
+/// <summary>
+///     Base class that provides timestamp fields to all Entities.
+/// </summary>
+public class Timestamps
```
TheTedder, line 7: I don't think this class is meant to be here anymore.

LeaderboardBackend/Models/Entities/NumericalMetric.cs
```diff
@@ -0,0 +1,96 @@
...
+	///     The minimum value of the `NumericalMetric`. The default is 0.
+	/// </summary>
+	[Required]
+	public long Min { get; set; }
+
+	/// <summary>
+	///     The maximum value of the `NumericalMetric`. The default is `Min` + 1.
+	/// </summary>
+	[Required]
+	public long Max { get; set; }
```
TheTedder, line 35: This should probably default to the integer limit. Otherwise, there are literally only two values it can even be by default.

LeaderboardBackend/Models/Entities/BaseEntity.cs
zysim: Real talk I don't know what I'm doing half the time with overriding these methods. So it should just be based on ID usually?

LeaderboardBackend/Models/Entities/BaseEntity.cs
zysim: Since there isn't an `Id` field here, I naturally couldn't match by ID. But as I've mentioned in Discord, I'm thinking about adding the `Id` field in here too, since all entities will have an ID column. And by having equality be ID-based, do you mean that it should only match by ID, or match by ID + other fields?

That being said; as you may be able to tell, each entity's `Equals` overrides match a varying number of fields. Not sure on how I should implement those either. It'd be great if you could take a look at those and advise too.

LeaderboardBackend/Models/Entities/BaseEntity.cs
zysim: Will implement this change in #128

LeaderboardBackend/Models/Entities/BaseEntity.cs
zysim: Also ye I'll implement this change in #128

LeaderboardBackend/Models/Entities/NumericalMetric.cs
zysim: Was simply following the LucidChart. But yeah sure I'll rename it + add a migration.

LeaderboardBackend/Models/Entities/NumericalMetric.cs
zysim: The latter. Once again looking at the LucidChart, these - IntegerMetrics I guess? - are a many-to-many with runs. EF Core conventions call for both entities to add an `ICollection` of the other. You should've already seen a `public List<NumericalMetric>? NumericalMetrics { get; set; }` line in `Run.cs`.

LeaderboardBackend/Models/Entities/Run.cs
zysim: It's [EF Core conventions on defining a many-to-many[(https://learn.microsoft.com/en-us/ef/core/modeling/relationships?tabs=fluent-api%2Cfluent-api-simple-key%2Csimple-key#many-to-many). With this, EF Core generates the migration with the appropriate join table and stuff.

LeaderboardBackend/Models/Entities/Timestamps.cs
zysim: Eh

Oops.

LeaderboardBackend/Models/Entities/NumericalMetric.cs
zysim: I'll swap to `int`s over `long`s, sure. This will clash with the DB design, which calls for `min` and `max` to be `double`s, heads up.

LeaderboardBackend/Controllers/CategoriesController.cs
zysim: Two options here:
1. The underlying service DB call throws an `ArgumentNullException` if the DB query turns up zilch. I just didn't add a catch block for that specifically, but I was lazy. That being said;
2. There _is_ another method I can call (e.g. `FirstOrDefault()`) which'll return `null` instead that I could use, tho that will involve me rewriting all services' Get calls, for consistency. So heads up on that.

LeaderboardBackend/Models/Entities/NumericalMetric.cs
zysim: Okay so the migration won't be trivial, as I've since learned. In fact I think I might need to write custom SQL, which I don't fully know how, and will need quite a bit of time to learn. Maybe it would be worth at this point to start a new PR?

zysim: There's no point in having to write a brand new migration with custom SQL just for this

LeaderboardBackend/Models/Entities/BaseEntity.cs
ayecptn: For Runs we have a Guid and for everthing else we have an Id as an unique identifier. So either we change Run to use a Id or we leave it as it is now.

zysim: I'm gonna change Runs to have longs as IDs

TheTedder:
Metrics are now contained in the runs themselves.

PR #130, opened 2022-10-11
Added Category for Runs

ayecptn:
- Added CategoryId to Run model
- Added GetCategory for RunController
- adapted tests
- Edited CreateRun to enter a CategoryId

### DIFF
--- a/LeaderboardBackend.Test/Judgements.cs
+++ b/LeaderboardBackend.Test/Judgements.cs
@@ -14,11 +14,11 @@ namespace LeaderboardBackend.Test;
 [TestFixture]
 internal class Judgements
 {
-	private static TestApiClient s_ApiClient = null!;
-	private static Leaderboard s_DefaultLeaderboard = null!;
-	private static TestApiFactory s_Factory = null!;
-	private static string? s_Jwt;
-	private static long s_CategoryId;
+	private static TestApiClient s_apiClient = null!;
+	private static Leaderboard s_defaultLeaderboard = null!;
+	private static TestApiFactory s_factory = null!;
+	private static string? s_jwt;
+	private static long s_categoryId;
 
 	private const string VALID_USERNAME = "Test";
 	private const string VALID_PASSWORD = "c00l_pAssword";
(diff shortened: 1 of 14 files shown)


LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -107,4 +110,25 @@ public async Task<ActionResult<List<Participation>>> GetParticipations(Guid id)
...
+		}
+
+		Category? category = await _categoryService.GetCategoryForRun(run);
+
+		if (category is null)
+		{
+			return NotFound("Category not found");
+		}
+
+		return Ok(category);
```
zysim, line 132: Like I guess I don't mind having this implementation, though I am considering having `_runService.GetRun()` be able to return its relationships. That way, you could do something like a (skipping error handling): 
```cs
Run? run = await _runService.GetRun(id, true); // second arg to get relationships
return run.Category;
```
which'll save us the second (honestly tiny) DB call.
@RageCage64 thoughts?

LeaderboardBackend/Services/ICategoryService.cs
```diff
@@ -6,4 +6,5 @@ public interface ICategoryService
 {
 	Task<Category?> GetCategory(long id);
 	Task CreateCategory(Category category);
+	Task<Category?> GetCategoryForRun(Run run);
```
zysim, line 9: And yeah consequently this may need to be dropped

LeaderboardBackend.Test/Judgements.cs
```diff
@@ -18,6 +18,7 @@ internal class Judgements
 	private static Leaderboard s_DefaultLeaderboard = null!;
 	private static TestApiFactory s_Factory = null!;
 	private static string? s_Jwt;
+	private static long s_CategoryId;
```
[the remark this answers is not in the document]
ayecptn: changed it for all Tests

LeaderboardBackend/Controllers/RunsController.cs
ayecptn: Could do it and does make sense this way. Just did it like this, because GetRun didn't return the relationships and I didn't think of your solution

zysim (approved):
Aight this is good. Thank you :relieved:

PR #132, opened 2023-01-29
Added single executable self-contained build

RageCage64:
These are the changes required for the staging deployment. This will build the app as a single executable. Also adds a production guard for loading the `.env` file, since in production we'll just read environment variables inside the docker container.

### DIFF
--- a/LeaderboardBackend/Program.cs
+++ b/LeaderboardBackend/Program.cs
@@ -19,12 +19,15 @@ using BCryptNet = BCrypt.Net.BCrypt;
 #region WebApplicationBuilder
 WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
 
-DotEnv.Load(options: new DotEnvOptions(
-	ignoreExceptions: false,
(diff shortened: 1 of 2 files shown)


LeaderboardBackend/Program.cs
```diff
@@ -19,12 +19,15 @@
 #region WebApplicationBuilder
 WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
 
-DotEnv.Load(options: new DotEnvOptions(
-	ignoreExceptions: false,
-	overwriteExistingVars: false,
-	envFilePaths: new[] { builder.Configuration["EnvPath"] },
-	trimValues: true
-));
+if (!builder.Environment.IsProduction())
```
zysim, line 22: I've merged https://github.com/leaderboardsgg/leaderboard-backend/pull/134, which doesn't use `DotEnv` anymore. This statement needs updating

zysim (changes requested):

PR #134, opened 2023-02-09
Strongly typed configuration + validation

Dalet:
Changes:
* Refactored configuration to use the Options pattern
* Added configuration for db migration (will be useful for the dev env deployment)
* Changed the dotenv nuget to add dotenv as a configuration provider (only in Development env)

## Options pattern
Going forward all new configuration should be strongly typed and validated.

The main benefit of using the [Options pattern](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-7.0) is configuration validation.

Note that here I've injected it as `IOptions<ConfigObject>` in services, but `ConfigObject` could also be directly injected by also adding it as a singleton:
```csharp
builder.Services.AddSingleton(services => services.GetRequiredService<IOptions<ConfigObject>>().Value);
```
Injecting as `IOptions<>` introduces a dependency on `Microsoft.Extensions.Options` in services. Let me know if you have an opinion about this.

I've removed direct env vars readings and made it rely the framework's configuration system instead.
By default ASP.NET already loads environment variables, [but with a specific format](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-7.0#naming-of-environment-variables) (see `example.env` for updated env var names)

## `POSTGRES_TEST_PORT` configuration
I plan on removing this config when I add [Testcontainers](https://dotnet.testcontainers.org/) to integration tests in a future PR. A test DB container will automatically be created and binded to a random available port.

I might have missed other configurations, let me know in that case.

### DIFF
--- a/.github/workflows/test-backend.yml
+++ b/.github/workflows/test-backend.yml
@@ -18,6 +18,6 @@ jobs:
         - name: Spin up containers
           run: docker compose up -d
         - name: Run Tests on In Memory Database
-          run: USE_IN_MEMORY_DB=true dotnet test LeaderboardBackend.Test
+          run: ApplicationContext__UseInMemoryDb=true dotnet test LeaderboardBackend.Test
         - name: Run Tests on Test Postgres Database
-          run: USE_IN_MEMORY_DB=false dotnet test LeaderboardBackend.Test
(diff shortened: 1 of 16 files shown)


zysim (approved):
This is neat. Thanks for this

PR #135, opened 2023-03-05
Run integration tests against a Postgres container (Testcontainers)

Dalet:
## Motivation
Integration tests against an in-memory db are not reliable: some SQL requests can work perfectly against the in-memory db and fail on the real database. This PR aims to make running the tests against a real database very simple (no more docker-compose).
## Changes summary
- Changed the README to mention the Docker dependency and how to run in-memory tests

### Test project:
- Made tests run against a temporary Postgres Docker container via the [Testcontainers library](https://dotnet.testcontainers.org/)
  - The container is created only once and automatically deleted at the end.
  - The database is reset by dropping the db and recreating it from an already seeded template DB.
- Changed the setup of some test classes to run only once for faster execution time.

### CI/build changes:
- Removed db-test from docker-compose because it is replaced by Testcontainers (and its associated `POSTGRES_TEST_PORT` env variable from `example.env`
- Made .env file completely optional (.csproj edits), and removed the step that creates it in the CI.
- Separated nuget restore, build, and test steps on the CI so it is easier to see where failure occurs.
- Removed in-memory tests (it's still possible to run them with this env variable: `INTEGRATION_TESTS_DB_BACKEND=InMemory`)

### New dependencies:
- [Testcontainers](https://dotnet.testcontainers.org/)
- Docker

## Performance
I've made resetting the database as fast as I could, but keep in mind resetting the db is slow, and often easily avoidable.

API requests in `[SetUp]` also take a lot of time if there are a lot of tests in the class, so they should be kept to a minimum as well.

### DIFF
--- a/LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
+++ b/LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
@@ -1,0 +1,106 @@
+using System;
+using System.Threading.Tasks;
+using DotNet.Testcontainers.Builders;
+using DotNet.Testcontainers.Configurations;
+using DotNet.Testcontainers.Containers;
+using LeaderboardBackend.Test.Fixtures;
+using Npgsql;
+using NUnit.Framework;
+
+// Fixtures apply to all tests in its namespace
+// It has no namespace on purpose, so that the fixture applies to all tests in this assembly
+
+[SetUpFixture] // https://docs.nunit.org/articles/nunit/writing-tests/attributes/setupfixture.html
+internal class PostgresDatabaseFixture
+{
+	public static TestcontainerDatabase? PostgresContainer { get; private set; }
+	public static bool HasCreatedTemplate { get; private set; } = false;
+
+	private static string TemplateDatabase => PostgresContainer?.Database + "_template";
+
+	[OneTimeSetUp]
+	public static async Task OneTimeSetup()
+	{
+		if (TestConfig.DatabaseBackend != DatabaseBackend.TestContainer)
+		{
+			return;
+		}
+
+		PostgresContainer = new ContainerBuilder<PostgreSqlTestcontainer>()
+			.WithDatabase(new PostgreSqlTestcontainerConfiguration
+			{
+				Database = "db",
+				Username = "dbuser",
+				Password = "weft5gst768sr",
+			})
+			.WithTmpfsMount("/var/lib/postgresql/data") // db files in-memory
+			.Build();
+		await PostgresContainer.StartAsync();
+	}
+
+	[OneTimeTearDown]
+	public static async Task OneTimeTearDown()
+	{
+		if (PostgresContainer is null)
+		{
+			return;
+		}
+
+		await PostgresContainer.DisposeAsync();
+	}
+
+	public static void CreateTemplateFromCurrentDb()
+	{
+		ThrowIfNotInitialized();
+
+		NpgsqlConnection.ClearAllPools(); // can't drop a DB if connections remain open
+		using NpgsqlDataSource conn = CreateConnectionToTemplate();
+		conn.CreateCommand(@$"
+			DROP DATABASE IF EXISTS {TemplateDatabase};
+			CREATE DATABASE {TemplateDatabase}
+				WITH TEMPLATE {PostgresContainer!.Database}
+				OWNER '{PostgresContainer!.Username}';
+			")
+			.ExecuteNonQuery();
+		HasCreatedTemplate = true;
+	}
+
+
+	// It is faster to recreate the db from an already seeded template
+	// compared to dropping the db and recreating it from scratch
+	public static void ResetDatabaseToTemplate()
+	{
+		ThrowIfNotInitialized();
+		if (!HasCreatedTemplate)
+		{
+			throw new InvalidOperationException("Database template has not been created.");
+		}
+		NpgsqlConnection.ClearAllPools(); // can't drop a DB if connections remain open
+		using NpgsqlDataSource conn = CreateConnectionToTemplate();
+		conn.CreateCommand(@$"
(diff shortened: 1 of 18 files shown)


LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		PostgresContainer = new ContainerBuilder<PostgreSqlTestcontainer>()
+			.WithDatabase(new PostgreSqlTestcontainerConfiguration
+			{
+				Database = "db",
+				Username = "dbuser",
+				Password = "weft5gst768sr",
+			})
+			.WithTmpfsMount("/var/lib/postgresql/data") // db files in-memory
```
RageCage64, line 36: Is this mount on the container or the system? Cause if it's on the system we'll need to make this work on Windows too.

RageCage64 (commented):
If we always have to do the same setup in every test fixture, can we make a base class? Maybe something like `APIResourceTestFixture`?

LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
Dalet: this is on the container

Dalet: Actually I'm not sure, but I tested it on Windows.

README.md
```diff
@@ -145,6 +145,8 @@ dotnet run --project LeaderboardBackend --urls https://localhost:7128 // dotnet
 The value provided to `--urls` has to match what's listed under `applicationUrl` in `LeaderboardBackend/Properties/launchSettings.json`. As of this writing, it's `https://localhost:7128`.
 
 #### Test the App
+Docker is required to run integration tests against a real database.  
+However, if you would like to run the tests with an in-memory database, you can set the following environment variable: `INTEGRATION_TESTS_DB_BACKEND=InMemory`.
```
zysim, line 149: Since you've mentioned that we can have situations where tests can work if we use an in-memory DB, but fail in a real one, do we want to mention that here? Maybe encourage contributors to always test with a real DB instance before requesting review?

LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
```diff
@@ -0,0 +1,106 @@
...
+
+	[OneTimeSetUp]
+	public static async Task OneTimeSetup()
+	{
+		if (TestConfig.DatabaseBackend != DatabaseBackend.TestContainer)
+		{
+			return;
+		}
+
+		PostgresContainer = new ContainerBuilder<PostgreSqlTestcontainer>()
```
zysim, line 29: Github Actions are warning that this constructor's obsolete. Mind fixing this accordingly?

zysim, line 29: Github Actions are warning that this constructor's obsolete. Mind fixing this accordingly?

LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
```diff
@@ -0,0 +1,106 @@
...
+
+		NpgsqlConnection.ClearAllPools(); // can't drop a DB if connections remain open
+		using NpgsqlDataSource conn = CreateConnectionToTemplate();
+		conn.CreateCommand(@$"
+			DROP DATABASE IF EXISTS {TemplateDatabase};
+			CREATE DATABASE {TemplateDatabase}
+				WITH TEMPLATE {PostgresContainer!.Database}
+				OWNER '{PostgresContainer!.Username}';
+			")
+			.ExecuteNonQuery();
```
zysim, line 64: Sorry but I'm having a hard time following this. So `CreateConnectionToTemplate()` returns a DataSource that's connected to `PostgresContainer`, but with its database set to "template1". But `conn.CreateCommand()` would create a database named "db_template", wouldn't it? If so, do we use "template1" anywhere?

zysim (changes requested):

Dalet:
> If we always have to do the same setup in every test fixture, can we make a base class? Maybe something like `APIResourceTestFixture`?
Absolutely, I think it's a good idea but outside the scope of the PR.

LeaderboardBackend.Test/Fixtures/PostgresDatabaseFixture.cs
Dalet: https://github.com/testcontainers/testcontainers-dotnet/issues/750#issuecomment-1412257694
It wasn't possible to fix it until the next version of Testcontainers would release... which [it finally did 4 hours ago](https://github.com/testcontainers/testcontainers-dotnet/releases/tag/3.0.0)!
Will fix in the coming days.

zysim: Ah okay. Sweet

zysim (approved):

RageCage64:
> > If we always have to do the same setup in every test fixture, can we make a base class? Maybe something like `APIResourceTestFixture`?
> 
> Absolutely, I think it's a good idea but outside the scope of the PR.

I personally think it's in scope since this is the PR that adds the setup that matches a clear potential abstraction. But I don't feel strongly enough to block anything.

zysim (approved):

zysim:
> > > If we always have to do the same setup in every test fixture, can we make a base class? Maybe something like `APIResourceTestFixture`?
> > 
> > 
> > Absolutely, I think it's a good idea but outside the scope of the PR.
> 
> I personally think it's in scope since this is the PR that adds the setup that matches a clear potential abstraction. But I don't feel strongly enough to block anything.

Let's just make that change in its own PR. Don't foresee any clashes with other PRs and stuff.

PR #139, opened 2023-04-15
Automatically generate OpenApi spec file + CI check

Dalet:
# Changes summary
- Fix `LeaderboardBackend.csproj` indendation according to the .editorconfig rules (2 spaces)
- Add a build step (.csproj line 60) that generates the `openapi.json` file using `dotnet tool run swagger`.
Since it needs to run the application to generate the file, it needs a valid configuration. So I added a `swagger-gen.env` file to be used with the CLI tool.
- Ensure `openapi.json` is checked out with LF line endings via `.gitattributes`, because the tool always generates with LF.
- Add a CI check in case the `openapi.json` changes were not committed.

Look at the latest commit separately to see the actual changes I made. The PR diff says I changed everything in the csproj because I fixed the indentation in the first commit.

# Motivation
- **Have an always up-to-date `openapi.json` file inside the repository to be able to generate an API client without running the server.**
- **Make sure the OpenApi spec changes are reviewed.**

Currently the frontend team has to run the server locally to access the file which is not ideal.

I explicitly want the file to be committed manually (and not by the CI) because the OpenApi spec changes should be manually reviewed (unintended breaking changes, unexpected object serialization...).

I named the file `openapi.json` because it is recommended here: https://swagger.io/specification/ ("Document Structure" section)

### DIFF
--- a/LeaderboardBackend/LeaderboardBackend.csproj
+++ b/LeaderboardBackend/LeaderboardBackend.csproj
@@ -49,4 +49,17 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
+    <OpenApiFile Include="openapi.json" />
+    <Content Remove="@(OpenApiFile)" />
+    <None Include="@(OpenApiFile)" />
+    <!-- required to be detected by the "fast up-to-date" feature of Visual Studio https://github.com/dotnet/project-system/blob/main/docs/up-to-date-check.md -->
+    <UpToDateCheckOutput Include="@(OpenApiFile)" /> 
+  </ItemGroup>
+
+  <Target Name="SwaggerPostBuildTarget" Condition="'$(DISABLE_OPENAPI_FILE_GEN)' != 'true'" AfterTargets="Build" Inputs="$(OutputPath)/$(AssemblyName).dll" Outputs="@(OpenApiFile)">
+    <Exec Command="dotnet tool restore"></Exec>
+    <Exec EnvironmentVariables="EnvPath=../swagger-gen.env;ASPNETCORE_ENVIRONMENT=development" Command="dotnet tool run swagger tofile --output &quot;@(OpenApiFile)&quot; &quot;$(OutputPath)/$(AssemblyName).dll&quot; v1"></Exec>
+  </Target>
+
 </Project>
--- a/LeaderboardBackend/LeaderboardBackend.csproj
+++ b/LeaderboardBackend/LeaderboardBackend.csproj
@@ -1,52 +1,52 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
-	<PropertyGroup>
-		<TargetFramework>net7.0</TargetFramework>
-		<Nullable>enable</Nullable>
-		<ImplicitUsings>enable</ImplicitUsings>
-		<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
-	</PropertyGroup>
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+  </PropertyGroup>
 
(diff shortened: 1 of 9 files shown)


zysim:
> Ensure openapi.json is checked out with LF line endings via .gitattributes, because the tool always generates with LF.
This is done solely for the Github action, right, or is there any other reason for this enforcement?

LeaderboardBackend/LeaderboardBackend.csproj
```diff
@@ -49,4 +49,17 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
+    <OpenApiFile Include="openapi.json" />
+    <Content Remove="@(OpenApiFile)" />
```
zysim, line 54: Where's `@(OpenApiFile)` defined?

swagger-gen.env
```diff
@@ -0,0 +1,5 @@
+# environment used for the openapi.json generation build step (see .csproj)
+
+ApplicationContext__UseInMemoryDb=true
+JWT__KEY=coolsecretkeywow
+JWT__ISSUER=leaderboards.gg
```
zysim, line 5: (Can't get the "comment on this file" action to work)

Can't we just use `.env`? And side question; the `JWT_*` values here don't have to agree with what's in `.env`, right?

LeaderboardBackend/LeaderboardBackend.csproj
```diff
@@ -1,52 +1,65 @@
...
+
+  <ItemGroup>
+    <None Include="../.env" Condition="Exists('../.env')">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+
+  <ItemGroup>
+    <OpenApiFile Include="openapi.json" />
+    <Content Remove="@(OpenApiFile)" />
```
zysim, line 54: What's the `@()` syntax? What does that resolve to?

swagger-gen.env
Dalet: `.env` is reserved for the developer's customized configuration, we can't use it.
This is a dummy configuration just to get the app to run without config validation errors basically.
Since Swagger generates from the actual API implementation, it needs to run Program.cs to setup the API to be able to detect the endpoints (I don't know the details).

LeaderboardBackend/LeaderboardBackend.csproj
Dalet: ["@ is for item, which typically is a group of files with attached metadatas under a name"](https://stackoverflow.com/a/6446506)
Items can be freely added with custom name. Here I added a "OpenApiFile" custom item on line 53.

Dalet: > > Ensure openapi.json is checked out with LF line endings via .gitattributes, because the tool always generates with LF.
> 
> This is done solely for the Github action, right, or is there any other reason for this enforcement?

On Windows the file may be checked out with CRLF line endings (depending on git config). 
Since the generation tool always outputs LF, git could say the file has changed even though only the line endings changed.

This is not required by the CI stuff, this is just QoL for developers.

Dalet:
> > Ensure openapi.json is checked out with LF line endings via .gitattributes, because the tool always generates with LF.
> 
> This is done solely for the Github action, right, or is there any other reason for this enforcement?

On Windows the file may be checked out with CRLF line endings (depending on git config). 
Since the generation tool always outputs LF, git could say the file has changed even though only the line endings changed.

This is not required by the CI stuff, this is just QoL for developers.

swagger-gen.env
Dalet: By the way if I had added the env variables to the `EnvironmentVariables` attribute [here](https://github.com/leaderboardsgg/leaderboard-backend/pull/139/files#diff-cd33d4cae6d4b44f2d39be061ba6f65da680c5b0c741b456743cdfb769a2719cR62), there would be no need for this file.

zysim: Let's keep `swagger-gen.env`. It's easier to read the task.

zysim (approved):

PR #140, opened 2023-04-23
Get leaderboard by slug + add Leaderboard and Category view models

### LINKED ISSUE #136: Feature Request: Enable fetching Leaderboard information by its unique `Slug`
I would like to propose that we add an API route to fetch the information for a given leaderboard by its unique `slug`. This will help the front end application more easily be able to request the correct data. 

See https://github.com/leaderboardsgg/leaderboard-site/issues/545 as to how it would be used

### Affected Functionality
Nothing should be effected by this change

### Other Relevant Issues
https://github.com/leaderboardsgg/leaderboard-site/issues/545

Dalet:
Closes #136

## Changes summary
- Added view models for `Leaderboard` and `Category`
- Added a way to get a leaderboard by its slug `GET /api/leaderboards/{slug}`
- Refactored the leaderboard tests with Bogus and FluentAssertions
- Added Bearer auth to the OpenAPI spec (also adds it to Swagger UI)

## Notes
I broke the current [ViewModel guideline](https://github.com/leaderboardsgg/leaderboard-backend/wiki/Style-Guide#viewmodels) to show what I think would be better (this wasn't possible before the project was upgraded to .NET 7):
- Each property should be `required`
- A ViewModel should have a static `MapFrom(TDomainEntity entity)` method.
  The `required` keyword ensures every property has been mapped. It doesn't work inside the constructor though so it needs to be a method.

### DIFF
--- a/LeaderboardBackend/Controllers/LeaderboardsController.cs
+++ b/LeaderboardBackend/Controllers/LeaderboardsController.cs
@@ -2,6 +2,7 @@ using LeaderboardBackend.Authorization;
 using LeaderboardBackend.Controllers.Annotations;
 using LeaderboardBackend.Models.Entities;
 using LeaderboardBackend.Models.Requests;
+using LeaderboardBackend.Models.ViewModels;
 using LeaderboardBackend.Services;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -28,7 +29,7 @@ public class LeaderboardsController : ControllerBase
 	[AllowAnonymous]
 	[ApiConventionMethod(typeof(Conventions), nameof(Conventions.GetAnon))]
 	[HttpGet("{id}")]
-	public async Task<ActionResult<Leaderboard>> GetLeaderboard(long id)
+	public async Task<ActionResult<LeaderboardViewModel>> GetLeaderboard(long id)
 	{
 		Leaderboard? leaderboard = await _leaderboardService.GetLeaderboard(id);
 
@@ -37,7 +38,7 @@ public class LeaderboardsController : ControllerBase
 			return NotFound();
 		}
 
-		return Ok(leaderboard);
+		return Ok(LeaderboardViewModel.MapFrom(leaderboard));
 	}
 
 	/// <summary>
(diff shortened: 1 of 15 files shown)


LeaderboardBackend/Controllers/LeaderboardsController.cs
```diff
@@ -37,24 +38,27 @@ public async Task<ActionResult<Leaderboard>> GetLeaderboard(long id)
...
 	///     The list of `Leaderboard`s was retrieved successfully. The result can be an empty
 	///     collection.
 	/// </response>
 	[AllowAnonymous]
 	[HttpGet]
 	[ProducesResponseType(StatusCodes.Status200OK)]
-	public async Task<ActionResult<List<Leaderboard>>> GetLeaderboards(
-		[FromQuery] long[] ids)
+	public async Task<ActionResult<List<LeaderboardViewModel>>> GetLeaderboards(
+		[FromQuery] long[] ids, [FromQuery] string? slug = null)
```
zysim, line 57: There can't ever be two leaderboards with the same slug, right? We could instead have a `GetLeaderboardBySlug` method maybe which returns a `LeaderboardViewModel` instead of a List of them. I'd imagine it'd be annoying for users to find a leaderboard by its slug to then have to get the only element out of the list every time.

I believe we can typehint the parameter in the REST method attribute, which should allow two methods with the "same" `[HttpGet("{id}")]`

(And if this works of course the tests will need to be changed, sorry)

LeaderboardBackend/Models/ViewModels/CategoryViewModel.cs
```diff
@@ -0,0 +1,49 @@
+using LeaderboardBackend.Models.Entities;
+
+namespace LeaderboardBackend.Models.ViewModels;
+
+/// <summary>
+///     Represents a `Category` tied to a `Leaderboard`.
+/// </summary>
+public record CategoryViewModel
```
zysim, line 8: Categories also have slugs. Could you please add them here too?

zysim (changes requested):

LeaderboardBackend/Models/Entities/Modship.cs
```diff
@@ -34,6 +34,7 @@ public class Modship
 	/// <summary>
 	///     Relationship model for `LeaderboardId`.
 	/// </summary>
+	[JsonIgnore]
 	public Leaderboard? Leaderboard { get; set; } = null!;
```
Dalet, line 38: This is because I wanted to remove all references to the Leaderboard entity from the `openapi.json` spec when I created the Leaderboard view models; and I was too lazy to create a Modship view model because we won't have moderators in MVP.

zysim: Fair. And there's definitely no point in creating the Modship view model anyway; no worries about "being lazy" about that :+1:

zysim: Could you add a comment there explaining the reason for it then? So we remember

LeaderboardBackend/Controllers/LeaderboardsController.cs
Dalet: Continuation on the issue: https://github.com/leaderboardsgg/leaderboard-backend/issues/136#issuecomment-1528681722

zysim (approved):

PR #145, opened 2023-05-14
Remove non-MVP entities

Dalet:
Removed code that is not relevant for MVP (judgements, participations, modships).

This should make future refactorings a bit less painful.


.husky/pre-commit
zysim, line 1: Why are the file perms changed here?

PR #146, opened 2023-05-14
Add run and user view models

### LINKED ISSUE #131: Task: Create UserViewModel
For obvious security reasons, we should return a UserViewModel to relevant requests.

Fields to remove from em are:
* Email
* Password
* ID

Dalet:
Based on PR https://github.com/leaderboardsgg/leaderboard-backend/pull/145

I left out some fields because we have no spec requiring them for now:
- User viewmodel
  - `About`
  - `Admin`
- Run viewmodel:
  - `PlayedOn`
  - `Status`

Closes #131

### DIFF
--- a/LeaderboardBackend/Models/ViewModels/RunViewModel.cs
+++ b/LeaderboardBackend/Models/ViewModels/RunViewModel.cs
@@ -1,0 +1,34 @@
+using LeaderboardBackend.Models.Entities;
+using NodaTime;
+
+namespace LeaderboardBackend.Models.ViewModels;
+
+public record RunViewModel
+{
+    /// <summary>
+    ///     The unique identifier of the `Run`.<br/>
+    ///     Generated on creation.
+    /// </summary>
+    public required Guid Id { get; set; }
+
+    /// <summary>
+    ///     The time the request was made at.
+    /// </summary>
+    public required Instant SubmittedAt { get; set; }
+
+    /// <summary>
+    /// 	The ID of the `Category` for `Run`.
+    /// </summary>
+    public required long CategoryId { get; set; }
+
+    public static RunViewModel MapFrom(Run run)
+    {
+        return new RunViewModel
+        {
+            Id = run.Id,
+            CategoryId = run.CategoryId,
+            SubmittedAt = run.SubmittedAt
+        };
+    }
+}
+
--- a/LeaderboardBackend/openapi.json
+++ b/LeaderboardBackend/openapi.json
@@ -462,12 +462,12 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/Run"
+                  "$ref": "#/components/schemas/RunViewModel"
                 }
               },
               "text/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/Run"
+                  "$ref": "#/components/schemas/RunViewModel"
                 }
               }
             }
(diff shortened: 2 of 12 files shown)


LeaderboardBackend/openapi.json
```diff
@@ -462,12 +462,12 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/Run"
+                  "$ref": "#/components/schemas/RunViewModel"
```
zysim, line 465: I wonder if we could rename entities for the generated spec, so that the viewmodels won't have that "ViewModel" suffix.

You think that's even worth it?

LeaderboardBackend/Models/ViewModels/RunViewModel.cs
```diff
@@ -0,0 +1,34 @@
+using LeaderboardBackend.Models.Entities;
+using NodaTime;
+
+namespace LeaderboardBackend.Models.ViewModels;
+
+public record RunViewModel
```
zysim, line 6: Pending discussion in Discord, but if we do keep `Status` in the parent model, we should return it here, too.

LeaderboardBackend/openapi.json
```diff
@@ -1429,12 +1376,22 @@
...
+          0,
+          1,
+          2,
+          3,
+          4
+        ],
+        "type": "integer",
+        "description": "The status of the `Run`.<br />\r\n    - 0: Created<br />\r\n    - 1: Submitted<br />\r\n    - 2: Pending<br />\r\n    - 3: Approved<br />\r\n    - 4: Rejected",
+        "format": "int32"
+      },
```
zysim, line 1390: Pending conclusion in Discord, but we may need to make sure this gets removed, as run status isn't part of the current DB spec.

LeaderboardBackend/openapi.json
Dalet: https://github.com/domaindrivendev/Swashbuckle.AspNetCore#customize-schema-ids
Looks easy to do but I can't say if it's worth it.

TheTedder: It's fine the way it is imo.

LeaderboardBackend/Models/ViewModels/RunViewModel.cs
TheTedder: There is no run approval mechanism as of right now. Newly created runs appear instantly. Runs can be deleted so there needs to be a mechanism for that, even if it's just marking them with a deleted flag. I would think the api does not return deleted runs by default. Maybe a query parameter can be used for that.

TheTedder: In other words, the `Status` field should not exist at this time. I recommend `IsDeleted` or `Deleted` for handling the deletion mechanism.

LeaderboardBackend/openapi.json
zysim: Yeah screw that then lol

LeaderboardBackend/Models/ViewModels/RunViewModel.cs
zysim: @Dalet yeah let's add an `IsDeleted` then I'll resolve this.

LeaderboardBackend/openapi.json
zysim: Yeah we should remove this. We don't have a run approval mechanism for now.

zysim (changes requested):

LeaderboardBackend/Models/ViewModels/RunViewModel.cs
zysim: ^ to be done in a future PR.

zysim (approved):

PR #151, opened 2023-06-02
Pin .NET SDK version

Dalet:
We had a workaround to make `dotnet format` work on the previous SDK version (7.0.203), but it's broken again on the latest version (7.0.302). So let's just pin it to an older version without problems (7.0.105) until it is resolved.

Previous .NET SDK versions can be downloaded here: https://dotnet.microsoft.com/en-us/download/dotnet/7.0 (multiple versions coexist on a system)

Also included a `launchSettings.json` fix (the ports were wrong)


LeaderboardBackend/Properties/launchSettings.json
```diff
@@ -1,14 +1,14 @@
...
+    "LeaderboardBackend": {
       "commandName": "Project",
       "launchBrowser": true,
       "launchUrl": "swagger",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
       "dotnetRunMessages": true,
-      "applicationUrl": "https://localhost:7128;http://localhost:5240"
+      "applicationUrl": "https://localhost:8080;http://localhost:8000"
```
zysim, line 11: Could you edit the readme, too? There's a line near the bottom where I wrote the URL out explicitly.

Dalet: Edited (the urls were actually not needed)

zysim (approved):

PR #162, opened 2023-06-25
Add EmailSender

### LINKED ISSUE #147: Task: Email sender service
Parent issue: https://github.com/leaderboardsgg/leaderboard-backend/issues/144

## Functional requirements:
- [ ] The server must be able to send email messages programmatically.
- [ ] Any email messages that the server sends should originate from a domain owned by the organization or a subdomain thereof.
- [ ] The email address that the server uses to send messages should be sent from an address that is clearly marked as "no-reply."
- [ ] The contents of email messages sent from the no-reply address may include content indicating that the recipient should not reply to email messages sent from said account.

## Technical requirements:
- [ ] Create an email sender service with a simple interface similar to: (feel free to deviate)
```csharp
interface IEmailSender
{
    Task SendEmailAsync(string recipientAddress, string subject, string htmlMessage);
}
```
I suggest using the [MailKit](https://github.com/jstedfast/MailKit#using-mailkit) library.

- [ ] `SendEmailAsync` cannot throw for "expected" errors (i.e: SMTP server connection failure must not throw).
- [ ] Add strongly-typed optional SMTP configuration. In `SendEmailAsync()`, if there is no configuration, just log a warning and exit the method.
- [ ] Log an Information level message (including the subject) when sending an email
- [ ] Log an Error when sending fails

### Future enhancements (maybe not for MVP)
- [ ] Use the [outbox pattern](https://en.wikipedia.org/wiki/Inbox_and_outbox_pattern) for reliability. Basically this means emails are added to a queue and processed (and retried) by a background job.
  - I suggest using [Hangfire](https://www.hangfire.io/) for this

## Affected Functionality
This is new functionality required for account management.

## Other relevant issues
- [Registration endpoint](https://github.com/leaderboardsgg/leaderboard-backend/issues/148)
- Send confirmation email endpoint
- Send recovery email endpoint

Dalet:
Closes #147

### DIFF
--- a/LeaderboardBackend.Test/Features/Emails/EmailSenderTests.cs
+++ b/LeaderboardBackend.Test/Features/Emails/EmailSenderTests.cs
@@ -33,7 +33,7 @@ public class EmailSenderTests
     }
 
     [Test]
-    public async Task SendEmailAsync_ShouldSendEmail()
+    public async Task EnqueueEmailAsync_ShouldSendEmail()
     {
         string expectedRecipientAddress = new Bogus.DataSets.Internet().Email();
         const string EXPECTED_SUBJECT = "Account recovery";
@@ -44,7 +44,7 @@ public class EmailSenderTests
                 It.IsAny<CancellationToken>(), It.IsAny<ITransferProgress>()))
             .Callback((MimeMessage m, CancellationToken _, ITransferProgress _) => sentMessage = m);
 
-        await _sut.SendEmailAsync(expectedRecipientAddress, EXPECTED_SUBJECT, EXPECTED_MESSAGE);
+        await _sut.EnqueueEmailAsync(expectedRecipientAddress, EXPECTED_SUBJECT, EXPECTED_MESSAGE);
 
         _smtpClientMock.Verify(x => x.SendAsync(It.IsAny<MimeMessage>(),
                 It.IsAny<CancellationToken>(), It.IsAny<ITransferProgress>()), Times.Once());
@@ -68,9 +68,9 @@ public class EmailSenderTests
     }
 
     [Test]
-    public async Task SendEmailAsync_ShouldConnectUsingConfig()
+    public async Task EnqueueEmailAsync_ShouldConnectUsingConfig()
     {
-        await SendEmailAsync(_sut);
+        await EnqueueEmailAsync(_sut);
 
         _smtpClientMock.Verify(x =>
             x.ConnectAsync(_config.Smtp!.Host, _config.Smtp.Port, _config.Smtp.UseSsl, It.IsAny<CancellationToken>()),
@@ -78,12 +78,12 @@ public class EmailSenderTests
     }
 
     [Test]
-    public async Task SendEmailAsync_ConfigWithCredentials_ShouldAuthenticate()
+    public async Task EnqueueEmailAsync_ConfigWithCredentials_ShouldAuthenticate()
     {
         _config.Smtp!.Username = "username";
         _config.Smtp.Password = "password";
 
-        await SendEmailAsync(_sut);
+        await EnqueueEmailAsync(_sut);
 
         _smtpClientMock.Verify(x =>
             x.AuthenticateAsync(_config.Smtp!.Username, _config.Smtp.Password, It.IsAny<CancellationToken>()),
@@ -91,12 +91,12 @@ public class EmailSenderTests
     }
 
     [Test]
-    public async Task SendEmailAsync_ConfigWithoutCredentials_ShouldNotAuthenticate()
+    public async Task EnqueueEmailAsync_ConfigWithoutCredentials_ShouldNotAuthenticate()
     {
         _config.Smtp!.Username = null;
         _config.Smtp.Password = null;
 
-        await SendEmailAsync(_sut);
+        await EnqueueEmailAsync(_sut);
 
         _smtpClientMock.Verify(x =>
             x.AuthenticateAsync(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<CancellationToken>()),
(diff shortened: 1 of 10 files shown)


LeaderboardBackend/Services/Impl/EmailSender.cs
```diff
@@ -0,0 +1,62 @@
...
+        _logger = logger;
+        _smtpClient = smtpClient;
+        _sender = new MailboxAddress(name: _config.SenderName, address: _config.SenderAddress);
+    }
+
+    public async Task SendEmailAsync(string recipientAddress, string subject, string htmlMessage)
+    {
+        if (_config.Smtp is null)
+        {
+            _logger.LogWarning("Can't send email (subject: \"{EmailSubject}\"), SMTP configuration is missing",
```
TheTedder, line 26: Should this maybe be an error instead of a warning? This seems like it would be a serious issue.

LeaderboardBackend/Services/Impl/EmailSender.cs
```diff
@@ -0,0 +1,62 @@
...
+        message.To.Add(new MailboxAddress(name: null, address: recipientAddress));
+        message.Subject = subject;
+
+        BodyBuilder bodyBuilder = new() { HtmlBody = htmlMessage };
+        message.Body = bodyBuilder.ToMessageBody();
+
+        try
+        {
+            await SendEmail(message);
+            _logger.LogInformation("Sent email (subject \"{EmailSubject}\")", subject);
```
TheTedder, line 42: I'm not sure this is worth logging because the subjects of the emails we send are going to be very generic, along the lines of "Recover Your Account" or "Confirm Your Account." It might be better to log from whatever method calls the email sending method so it can give a helpful log message like "sent recoery email to account ace5bf6885baf6" instead.

LeaderboardBackend/Services/Impl/EmailSender.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+        catch (Exception exception)
+        {
+            _logger.LogError(exception, "Caught exception when sending email (subject: \"{EmailSubject}\")", subject);
+        }
+    }
+
+    private async Task SendEmail(MimeMessage message)
+    {
+        await _smtpClient.ConnectAsync(_config.Smtp!.Host, _config.Smtp.Port, _config.Smtp.UseSsl);
```
TheTedder, line 52: Is it a good idea to keep this connection open or just connect ad hoc like this? I am a bit ignorant on best practices here.

LeaderboardBackend.Test/Features/Emails/EmailSenderTests.cs
TheTedder, line 1: Please fix the github warnings or configure them to be less strict.

TheTedder (commented):
Looking good—just a few comments here and there.

LeaderboardBackend/Services/Impl/EmailSender.cs
Dalet: changed to error

LeaderboardBackend/Services/Impl/EmailSender.cs
Dalet: a single SMTP instance is now reused as talked about on Discord

LeaderboardBackend/Services/Impl/EmailSender.cs
```diff
@@ -0,0 +1,83 @@
...
+    public EmailSender(IOptions<EmailSenderConfig> config, ILogger<EmailSender> logger, ISmtpClient smtpClient)
+    {
+        _config = config.Value;
+        _logger = logger;
+        _smtpClient = smtpClient;
+        _sender = new MailboxAddress(name: _config.SenderName, address: _config.SenderAddress);
+        _semaphore = new SemaphoreSlim(1);
+    }
+
+    public async Task SendEmailAsync(string recipientAddress, string subject, string htmlMessage)
```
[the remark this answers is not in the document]
Dalet: I doubt we will need this. Better add it when we actually need it.

The process that triggered the email and the queuing will be within the same DB transaction.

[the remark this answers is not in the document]
TheTedder: Yes but we need to know if our email service is failing to send messages. That's a big deal. Say the email service encounters a problem and for whatever reason emails are not being sent. That means nobody will be able to register or recover their account. This method needs to report whether the email was successfully queued for sending so that we can potentially detect this error and investigate it.

zysim:
@Dalet I've rebased this with the dotnet version updates./

Dalet:
https://discord.com/channels/856595632030613524/1131543204957790348
This discussion led to requiring the following changes:

- let exceptions bubble up
- rename the method to `EnqueueEmail` to be in line with the future queue implementation

LeaderboardBackend/Services/Impl/EmailSender.cs
```diff
@@ -0,0 +1,82 @@
...
+        message.Body = bodyBuilder.ToMessageBody();
+
+        await _semaphore.WaitAsync();
+        try
+        {
+            await SendEmail(message);
+        }
+        catch (Exception exception)
+        {
+            _logger.LogError(exception, "Caught exception when sending email (subject: \"{EmailSubject}\")", subject);
```
TheTedder, line 47: Would it be better to log the exception instead of the email subject? If we do that, we just have to make sure the recipient address isn't logged.

TheTedder: Alternatively, log nothing and let the caller decide whether to catch the exception and log or not.

Dalet: (the exception is logged here)

I have no idea if the mailing library can display the email in its exceptions. I'm gonna remove the log for now.

TheTedder: Uncle Ted is going blind it seems.

TheTedder (approved):

PR #164, opened 2023-06-28
LBGG: Auto update API

erunks:
## What
In order to keep the frontend API in sync better with the changes being made here, this PR adds 1/2 of the workflow actions that will trigger the workflow which automatically creates a PR with all the API changes on https://github.com/leaderboardsgg/leaderboard-site.

## Requirements
Needs the following things to run:
  - repo secret `WORKFLOW_ACCESS_TOKEN`
    - this is what allows the `actions/github-script` to actually invoke the partner action on the frontend

## Changes
- Added:
  - new `trigger-api-update` workflow to trigger the `update-api` workflow on the frontend
  - some helpful bash scripts for starting/stopping the backend easily

## Link to Issue

Closes https://github.com/leaderboardsgg/leaderboard-site/issues/559

### Steps for testing

- Successful run on my fork of the backend repo: https://github.com/erunks/leaderboard-backend/actions/runs/5404668272
- Successful trigger on my fork of the frontend repo: https://github.com/erunks/leaderboard-site/actions/runs/5404338702
  - PR that was spun off from the frontend workflow: https://github.com/erunks/leaderboard-site/pull/3


TheTedder:
I'm not sure this is necessary. As long as we're in communication, this shouldn't be an issue.

scripts/start.sh
```diff
@@ -0,0 +1,6 @@
+dotnet tool restore
+cd LeaderboardBackend
+dotnet ef database update
+cd ../
+docker compose up -d
+dotnet run --project LeaderboardBackend --urls https://localhost:7128
```
Dalet, line 6: the `--urls` param is not necessary (iirc it didn't even work on my machine)

Dalet (changes requested):
Check that `openapi.json` was changed before triggering the other repo's workflow

scripts/start.sh
```diff
@@ -0,0 +1,6 @@
+dotnet tool restore
+cd LeaderboardBackend
+dotnet ef database update
+cd ../
+docker compose up -d
```
Dalet, line 5: `dotnet ef database update` should be after `docker compose up -d` since you need the database to be up to migrate it

zysim: Also you should be able to do `dotnet ef database update --project LeaderboardBackend` to skip having to `cd` into `LeaderboardBackend` and back

erunks: Added in those changes

zysim (approved):

Issue #149, opened 2023-05-21
Task: Login endpoint


Dalet:
Parent issue: https://github.com/leaderboardsgg/leaderboard-backend/issues/144

`POST /login`

Request body fields:
- email: string
- password: string

Successful response body fields:
- token: string

- [x] If the client does not supply both an email and a password in the request body, the server must respond with a `422 UNPROCESSABLE CONTENT` status code.
- [x] If the client supplies an email address that does not correspond to an existing user account, the server must respond with a `404 NOT FOUND` status code.
- [x] If the client supplies an email adresss that does correspond to an existing user account but the password does not match said user's password, the server must respond with a `401 UNAUTHORIZED` status code.
- [x] If the client supplies an email address that corresponds to an existing user account and a correct password that matches the encrypted password for said user whose role is `BANNED`, the server must return a `403 FORBIDDEN` status code.
- [x] Upon a successful login, the server must return a `200 OK` status code as well as a JSON web token in the response body.

### Affected Functionality
Login on front-end will break because of the URI change

### Other Relevant Issues
- [Give user feedback during account creation/login](https://github.com/leaderboardsgg/leaderboard-site/issues/563)

TheTedder:
Is this issue complete? If so, close it as completed.

PR #169, opened 2023-07-24
Login Endpoint

zysim:
Solves: #149


LeaderboardBackend/Program.cs
```diff
@@ -262,6 +262,11 @@
     options.InvalidModelStateResponseFactory = context =>
     {
         ValidationProblemDetails problemDetails = new(context.ModelState);
+        if (problemDetails.Errors.ContainsKey("$"))
+        {
+            return new BadRequestObjectResult(problemDetails);
+        }
```
TheTedder, line 268: This looks like a hack. Can you put a code comment here explaining what the purpose of this is?

LeaderboardBackend.Test/Features/Users/LoginTests.cs
```diff
@@ -0,0 +1,126 @@
...
+    private const string VALID_EMAIL = "valid@user.com";
+    private const string BANNED_EMAIL = "banned@user.com";
+    private const string VALID_PASSWORD = "P4ssword";
+
+    [OneTimeSetUp]
+    public void Init()
+    {
+        s_factory.ResetDatabase();
+
+        // Swap to creating users via the UserService instead of calling the DB, once
```
TheTedder, line 29: Mark this as `TODO:` please.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        // it has the ability to change a user's roles.
+        using IServiceScope s = s_factory.Services.CreateScope();
+        ApplicationContext dbContext = s.ServiceProvider.GetRequiredService<ApplicationContext>();
+        dbContext.Users.AddRange(new[]
+        {
+            new User{
+                Email = VALID_EMAIL,
+                Password = BCrypt.Net.BCrypt.EnhancedHashPassword(VALID_PASSWORD),
+                Username = "Test User",
```
TheTedder, line 38: Use valid data for testing. This is not a valid username. "Banned User" is not a valid username either. I suggest Test_User and Banned_User instead.

LeaderboardBackend.Test/consts.cs
```diff
@@ -0,0 +1,7 @@
+namespace LeaderboardBackend.Test;
+
+public static class Routes
+{
+    public static string s_login = "/login";
+    public static string s_register = "/account/register";
```
TheTedder, line 6: Make these variables `readonly`.

LeaderboardBackend/Program.cs
zysim: Oh yeah this is. I can't figure out a more elegant solution, so I did this.

I'll add a comment here, sure

TheTedder: Thank you. This is the kind of thing we're going to come back to in two months and be scratching our heads over.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using LeaderboardBackend.Test.Lib;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.DependencyInjection;
+using NUnit.Framework;
+
+namespace LeaderboardBackend.Test.Features.Users;
+
+public class LoginTests : IntegrationTestsBase
+{
+    private const string LOGIN_URI = "/login";
```
TheTedder, line 19: Use the constants you defined in consts.cs.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
```diff
@@ -0,0 +1,126 @@
...
+            Email = TestInitCommonFields.Admin.Email,
+            Password = TestInitCommonFields.Admin.Password,
+        };
+
+        HttpResponseMessage res = await Client.PostAsJsonAsync(LOGIN_URI, request);
+
+        res.Should().HaveStatusCode(HttpStatusCode.OK);
+        LoginResponse? content = await res.Content.ReadFromJsonAsync<LoginResponse>();
+        content.Should().NotBeNull();
+        content!.Token.Should().MatchRegex(@"^[A-Za-z0-9-_=]{36}\.[A-Za-z0-9-_=]{190}\.[A-Za-z0-9-_=]{43}$");
```
Dalet, line 64: Use this to check if it is a JWT `Jwt.SecurityTokenHandler.CanReadToken()`

Or even better, validate it: https://github.com/leaderboardsgg/leaderboard-backend/blob/48e31ccfb865b6f8e3f02c1fdb3d2a61b7de8156/LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs#L102

Dalet (changes requested):

LeaderboardBackend/Program.cs
```diff
@@ -262,6 +262,16 @@
     options.InvalidModelStateResponseFactory = context =>
     {
         ValidationProblemDetails problemDetails = new(context.ModelState);
+
+        // As of this writing, we want our custom user registration validation rules to
+        // return with a 429, with 400 for the rest. This is a hack that leverages
```
TheTedder, line 267: This should say 422 and not 429.

TheTedder (changes requested):

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -74,4 +74,60 @@ public async Task<ActionResult<UserViewModel>> Register([FromBody] RegisterReque
...
+    [AllowAnonymous]
+    [HttpPost("/login")]
+    [ApiConventionMethod(typeof(Conventions), nameof(Conventions.PostAnon))]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+#pragma warning disable CS1573 // Hides warning for not having authService in the XML comment above
+    public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest request, [FromServices] IAuthService authService)
+    {
```
Dalet, line 110: business logic should be in a service class

LeaderboardBackend.Test/Features/Users/LoginTests.cs
```diff
@@ -0,0 +1,126 @@
...
+    [TestCase("ee", VALID_PASSWORD, "EmailValidator", null, Description = "Null email + valid password")]
+    [TestCase(VALID_EMAIL, "ff", null, UserPasswordRule.PASSWORD_FORMAT, Description = "Valid email + null password")]
+    public async Task Login_InvalidRequest_Returns422UnprocessableEntity(
+        string? email,
+        string? password,
+        string? emailErrorCode,
+        string? passwordErrorCode)
+    {
+        // We should figure out how to have an UnvalidatedLoginRequest w/ null fields and a
+        // LoginRequest with non-null fields, so we don't have to force null coalescence - zysim
```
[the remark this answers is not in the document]
zysim: I was deciding between doing just that, and what I've done in this PR. I feel that manual JSON in the test case is less readable, after trying it out.

I'll just remove the comment.

LeaderboardBackend/Models/Requests/UserRequests.cs
```diff
@@ -93,6 +100,16 @@ public record RegisterRequest
     public required string Password { get; set; }
 }
 
+public class LoginRequestValidator : AbstractValidator<LoginRequest>
+{
+    public LoginRequestValidator()
+    {
+        // NotNull() call needed to get the error message that NotNull creates
+        RuleFor(x => x.Email).Cascade(CascadeMode.Stop).NotNull().EmailAddress();
+        RuleFor(x => x.Password).UserPassword();
```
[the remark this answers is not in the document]
zysim: ...oh yeah it isn't even necessary now, is it lol

zysim:
Actually; should we even bother with 422 in the first place? We could just do 400 for any malformed, empty, or otherwise invalid inputs.

LeaderboardBackend/Models/Requests/UserRequests.cs
```diff
@@ -93,6 +100,16 @@ public record RegisterRequest
     public required string Password { get; set; }
 }
 
+public class LoginRequestValidator : AbstractValidator<LoginRequest>
+{
+    public LoginRequestValidator()
+    {
+        // NotNull() call needed to get the error message that NotNull creates
```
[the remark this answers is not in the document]
zysim: On a tangent; I wonder how necessary having these return a 422 on failure is. Started a thread [below](https://github.com/leaderboardsgg/leaderboard-backend/pull/169#issuecomment-1669290100).

LeaderboardBackend/Models/Requests/UserRequests.cs
```diff
@@ -93,6 +100,16 @@ public record RegisterRequest
...
+        RuleFor(x => x.Email).Cascade(CascadeMode.Stop).NotNull().EmailAddress();
```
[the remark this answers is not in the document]
zysim: Huh. Swear I needed it at the time.

LeaderboardBackend/Program.cs
```diff
@@ -262,6 +262,16 @@
...
+        // FluentValidation's default error object shape to achieve that effect.
+        // If anyone knows of a better solution, please update this, thanks. - zysim
+        if (problemDetails.Errors.ContainsKey("$"))
```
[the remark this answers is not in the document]
TheTedder: Good suggestion. @zysim You should put this explanation in the code comment so we don't forget why we did this.

TheTedder:
> Actually; should we even bother with 422 in the first place? We could just do 400 for any malformed, empty, or otherwise invalid inputs.

I think we're reserving 400 for invalid syntax; i.e., valid JSON never returns 400.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
TheTedder: What's the point of doing `private const string LOGIN_URI = Routes.LOGIN;`? Just use Routes.LOGIN directly.

LeaderboardBackend.Test/consts.cs
```diff
@@ -0,0 +1,7 @@
+namespace LeaderboardBackend.Test;
+
+public static class Routes
+{
+    public const string LOGIN = "/login";
+    public const string REGISTER = "/account/register";
```
TheTedder, line 6: The use of `const` should be reserved for values that can never change, such as mathematical constants.

TheTedder (changes requested):

LeaderboardBackend/Program.cs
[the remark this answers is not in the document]
zysim: I feel that the current comment is enough.

LeaderboardBackend.Test/consts.cs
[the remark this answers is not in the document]
zysim: I imagine this file to hold global constants, which is why I named it so. Could rename it, but I'm not sure what.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
zysim: Why? Because I'm a genius.

(Changing)

LeaderboardBackend.Test/consts.cs
zysim: Don't see why we can only use `const` that way. It seems more like a philosophical decision. Also, removing it will prevent us from e.g. using them as attribute arguments (it's a small example, but it's something to keep in mind)

LeaderboardBackend/Controllers/AccountController.cs
zysim: I've made a new method `GetUserByEmailAndPassword` as I'd need to change the function signature, and the existing method already has use in the `Me()` action.

LeaderboardBackend.Test/consts.cs
TheTedder: From Microsoft's C# guide:

> Don’t create a constant to represent information that you expect to change at any time. For example, don’t use a constant field to store the price of a service, a product version number, or the brand name of a company. These values can change over time, and because compilers propagate constants, other code compiled with your libraries will have to be recompiled to see the changes. See also the [readonly](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly) keyword.

Endpoints can potentially change in the future, therefore it is not suitable to use `const` for them.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
```diff
@@ -0,0 +1,130 @@
...
+using Microsoft.IdentityModel.Tokens;
+using NUnit.Framework;
+
+namespace LeaderboardBackend.Test.Features.Users;
+
+public class LoginTests : IntegrationTestsBase
+{
+    private const string VALID_EMAIL = "valid@user.com";
+    private const string BANNED_EMAIL = "banned@user.com";
+    private const string VALID_PASSWORD = "P4ssword";
```
TheTedder, line 24: Same as elsewhere, reserve the use of `const` for unchanging constants. Use `readonly` instead. There is no performance benefit to marking variables as constant in C#.

LeaderboardBackend/Program.cs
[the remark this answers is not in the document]
TheTedder: Please add something along the lines of "The name of a JSON syntax error is always in the form of a relative path that begins with '$'."

LeaderboardBackend/Models/Requests/UserRequests.cs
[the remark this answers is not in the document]
TheTedder: Either way is fine but in my opinion it is better to validate that the email provided is a well-formed email. Being able to return two different errors for "what you entered was not a valid email" and "the email you entered didn't match any users" leads to a better user experience.

LeaderboardBackend/Program.cs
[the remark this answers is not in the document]
zysim: Sorry, I was misunderstanding things, which is why I thought there wasn't a need at first.

LeaderboardBackend/Models/Requests/UserRequests.cs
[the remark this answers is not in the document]
zysim: Fair. I'll keep the validator

LeaderboardBackend.Test/consts.cs
zysim: The More You Know 🌈 🌟

Thanks for letting me know. Changing them to be `static readonly`.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
zysim: Performance wasn't my concern here, but the fact that the `TestCase` attributes don't like it when their arguments aren't `const`s. Swapping them to `readonly` makes even declaring these fields a bit useless, IMO, as they get very little use outside of the attribute args.

LeaderboardBackend.Test/consts.cs
Dalet: While the advice is generally correct, this a test project so the problem it avoids does not apply in this context.

Dalet: You can make the const `internal` to prevent the issue entirely.

LeaderboardBackend.Test/Features/Users/LoginTests.cs
Dalet: That "best practice" is only for public constants.
This is fine.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -74,4 +72,59 @@ public async Task<ActionResult<UserViewModel>> Register([FromBody] RegisterReque
...
+                    return NotFound();
+                }
+
+                if (errors.Banned)
+                {
+                    return Forbid();
+                }
+
+                // errors.WrongPassword implicitly true
+                return Unauthorized();
```
zysim, line 126: I don't like what I've done here, but I think the alternatives are worse, so 🤷

LeaderboardBackend.Test/consts.cs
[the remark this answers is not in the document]
TheTedder: Consts is fine I suppose but Routes may be better. If would be cool if there were a way to just get these from the controller somehow but this is fine for now.

LeaderboardBackend/Controllers/AccountController.cs
TheTedder: You could use [pattern matching](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#multiple-inputs).

TheTedder (commented):
Looking good overall. Just one or two things to tweak. Since @Dalet didn't agree about not using consts, I don't mind if you want to revert them.

LeaderboardBackend.Test/consts.cs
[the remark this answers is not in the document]
Dalet: +1 for `Consts.cs`
Having routes as constants in tests is valuable, tests should break when the endpoint path is changed.

LeaderboardBackend/Controllers/AccountController.cs
zysim: Discussed it with @Dalet and I just need to leverage the `OneOf` library better. I'll push a commit soon showing what I mean.

LeaderboardBackend.Test/consts.cs
zysim: @TheTedder I'll revert the members to consts. And I'll also make the class itself internal, instead of the members themselves, since that makes more sense IMO.

LeaderboardBackend/Controllers/AccountController.cs
zysim: https://github.com/leaderboardsgg/leaderboard-backend/pull/169/commits/53064f1e1ec758f7f5d4fe45415e51130d7fc492

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -74,4 +72,59 @@ public async Task<ActionResult<UserViewModel>> Register([FromBody] RegisterReque
...
+    ///       - **EmailValidator**: Invalid email format
+    /// </response>
+    [AllowAnonymous]
+    [HttpPost("/login")]
+    [ApiConventionMethod(typeof(Conventions), nameof(Conventions.PostAnon))]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    [ProducesResponseType(StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest request, [FromServices] IAuthService authService)
```
Dalet, line 107: these can be removed now
- `[FromServices] IAuthService authService`
- `/// <param name="authService">IAuthService injection.</param>`

Dalet (approved):

TheTedder (approved):

Issue #154, opened 2023-06-24
Task: Resend account confirmation link endpoint


Dalet:
Parent issue: #153

```[tasklist]
### Blocked by
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/147
```

Endpoint: `POST /account/confirm`
Requires authentication.

## Responses
- [x] If the client does not supply a valid session token, the server must respond with a `401 UNAUTHORIZED` status code.
- [x] If the session token provided was created on behalf of a user whose role is not `REGISTERED`, the server must respond with a `409 CONFLICT` status code.
- [x] If the client supplies a valid session token that was created on behalf of a user whose role is `REGISTERED`, the server must respond with a `200 OK` status code and initiate the confirmation link generation procedure. (See below.) The server may respond to the request before the email is sent to prevent a timeout.
- [ ] The server may impose a limit on how often a client can call this endpoint. If the client calls this endpoint too many times within a given timespan, the server may respond with a `429 TOO MANY REQUESTS` status code.

## Procedure 
Inputs:
* user ID: GUID

### Requirements
- [x] A new confirmation record must be created in the database, the `UserId` field must be the ID of the authenticated user.
- [x] An email must be sent to the email address supplied in the email field of the user specified by the `UserId` field of the confirmation.
  - TODO: Determine the exact contents of the email.
  - [x] The email must contain a link to the account confirmation page.
    - [x] The account confirmation page link must have the path `/confirm-account?code={id}` where `{id}` is the newly-created confirmation's ID in Base64 encoding .
- [x] The confirmation token must expire 1 hour after creation

PR #172, opened 2023-08-03
Resend Confirmation Email

zysim:
Closes: #154


Dalet (changes requested):

LeaderboardBackend/Models/Entities/ApplicationContext.cs
```diff
@@ -19,6 +19,7 @@ public ApplicationContext(DbContextOptions<ApplicationContext> options)
         : base(options) { }
 
     public DbSet<Category> Categories { get; set; } = null!;
+    public DbSet<Confirmation> Confirmations { get; set; } = null!;
```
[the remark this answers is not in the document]
zysim: Should the type also be `UserConfirmation` too, then?

LeaderboardBackend/Models/Entities/User.cs
```diff
@@ -76,6 +76,11 @@ public class User
     [Required]
     public UserRole Role { get; set; } = UserRole.Registered;
 
+    /// <summary>
+    ///     A collection of `Confirmation` entities for the `User`.
+    /// </summary>
+    public List<Confirmation>? Confirmations { get; set; }
```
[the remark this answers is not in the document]
zysim: I was always under the impression that I need to have the navigation in order for the migration file to be generated correctly. Didn't know I could omit that.

LeaderboardBackend/Models/Entities/ApplicationContext.cs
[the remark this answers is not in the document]
Dalet: Yes ofc

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+[TestFixture]
+public class AccountConfirmationTests : IntegrationTestsBase
+{
+    private const string RESEND_CONFIRMATION_URI = "/account/confirm";
+    private IServiceScope _scope = null!;
+    private IAuthService _authService = null!;
+    private IUserService _userService = null!;
+
+    [OneTimeSetUp]
```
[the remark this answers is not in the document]
zysim: Happy to change it, but in the meantime, but could you explain why it could be bad? I can't really see why.

Also, I'm rereading [the docs on scoped services](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#scoped), and it warns to not resolve them from a singleton. Since we're resolving them from `s_factory`, which is a static member of a parent class, wouldn't that make it one?

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -130,4 +132,74 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+        string? email = authService.GetEmailFromClaims(HttpContext.User);
+
+        if (email is null)
+        {
+            return Unauthorized();
+        }
+
+        User? user = await _userService.GetUserByEmail(email);
+
+        if (user is null)
```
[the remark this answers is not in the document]
zysim: Yeah, that makes more sense.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            Email = "email1",
+            Password = "password",
+            Username = "username1",
+        });
+        string token = _authService.GenerateJSONWebToken(result.AsT0);
+
+        Client.DefaultRequestHeaders.Authorization = AuthenticationHeaderValue.Parse($"Bearer {token}");
+        HttpResponseMessage res = await Client.PostAsync(RESEND_CONFIRMATION_URI, null);
+        res.Should().HaveStatusCode(HttpStatusCode.OK);
```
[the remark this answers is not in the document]
zysim: Gonna make a separate test file for this.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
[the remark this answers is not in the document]
Dalet: Tests could share state and affect each others unexpectedly.
For example the DbContext's change tracker is shared in one scope, so one test could save another test's entity changes when calling .SaveChanges()

LeaderboardBackend/Controllers/AccountController.cs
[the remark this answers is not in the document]
Dalet: Ah I didn't think you would change it to get user id now.
Since it really shouldn't happen, it should throw an exception and not be documented.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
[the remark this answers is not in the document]
zysim: I see. I haven't considered that.

LeaderboardBackend/Controllers/AccountController.cs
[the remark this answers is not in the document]
zysim: Returning 500 instead of throwing an exception, so we can hide the stack trace and message to clients.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -114,4 +116,60 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+            await confirmationService.CreateConfirmation(user);
+            return Ok();
+        }
+
+        return errors.Match(
+            badCredentials => Unauthorized(),
+            // Shouldn't be possible; throw 500
+            notFound => StatusCode(StatusCodes.Status500InternalServerError),
+            badRole => Conflict()
+        );
```
zysim, line 173: I either keep this TryPick/Match pattern, or do this instead, if we want everything under a single Match: ```cs
return result.Match(
    user =>
    {
        var _ = confirmationService.CreateConfirmation(user).Result;
        return Ok();
    },
    badCredentials => Unauthorized(),
    // Shouldn't be possible; throw 500
    notFound => StatusCode(StatusCodes.Status500InternalServerError),
    badRole => Conflict()
);
```
`Match` doesn't like async functions

File: LeaderboardBackend/Models/Entities/ApplicationContext.cs (hunk 1 above)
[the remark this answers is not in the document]
Comment (TheTedder) [reply LeaderboardBackend/Models/Entities/ApplicationContext.cs]: Change the name of the file too.

File: LeaderboardBackend/Controllers/AccountController.cs [hunk 7]
```diff
@@ -114,4 +116,60 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+        if (result.TryPickT0(out User user, out OneOf<BadCredentials, UserNotFound, BadRole> errors))
+        {
+            await confirmationService.CreateConfirmation(user);
+            return Ok();
+        }
+
+        return errors.Match(
+            badCredentials => Unauthorized(),
+            // Shouldn't be possible; throw 500
+            notFound => StatusCode(StatusCodes.Status500InternalServerError),
```
TheTedder, line 171: Status code 500 should be reserved for actual server errors. Return 401 instead.

TheTedder (changes requested):

LeaderboardBackend/Controllers/AccountController.cs
TheTedder: It's fine the way it is.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -114,4 +116,60 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+    ///     The request doesn't contain a valid session token.
+    /// </response>
+    /// <response code="409">
+    ///     A `User` with the specified username or email already exists.<br/><br/>
+    ///     Validation error codes by property:
+    ///     - **Username**:
+    ///       - **UsernameTaken**: the username is already in use
+    ///     - **Email**:
+    ///       - **EmailAlreadyUsed**: the email is already in use
+    /// </response>
```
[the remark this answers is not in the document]
zysim: Oh lol I copied this block from the registration action and forgot to edit it.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -114,4 +116,51 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
     }
+#pragma warning restore CS1573
+
+    /// <summary>
+    ///     Resends the account confirmation link.
+    /// </summary>
+    /// <param name="authService">IAuthService dependency.</param>
+    /// <param name="confirmationService">IAccountConfirmationService dependency.</param>
+    /// <param name="emailSender">EmailSender dependency.</param>
+    /// <response code="200">The request was sent successfully.</response>
```
TheTedder, line 127: Should this be more descriptive? You could write something like "A new confirmation link was generated."

LeaderboardBackend/Models/Entities/AccountConfirmation.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    ///     The time this `AccountConfirmation` is used.
+    /// </summary>
+    public Instant? UsedAt { get; set; }
+
+    /// <summary>
+    ///     The time this `AccountConfirmation` expires, after which the associated `User` must
+    ///     request for another confirmation email.
+    /// </summary>
+    public Instant ExpiresAt { get; set; } = SystemClock.Instance.GetCurrentInstant() + Duration.FromHours(1);
```
TheTedder, line 42: Move this logic out of the entity definition.

LeaderboardBackend/Models/Entities/AccountConfirmation.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    ///     Relationship model for `UserId`.
+    /// </summary>
+    public User User { get; set; } = null!;
+
+    /// <summary>
+    ///     The time this `AccountConfirmation` is created.
+    /// </summary>
+    public Instant CreatedAt { get; set; } = SystemClock.Instance.GetCurrentInstant();
```
TheTedder, line 31: Move this out of entity definition.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,48 @@
...
+        _applicationContext = applicationContext;
+        _emailSender = emailSender;
+    }
+
+    public async Task<AccountConfirmation?> GetConfirmationById(Guid id)
+    {
+        return await _applicationContext.AccountConfirmations.FindAsync(id);
+    }
+
+    public async Task<AccountConfirmation> CreateConfirmation(User user, CancellationToken token = default)
```
TheTedder, line 21: This function name should be changed so that it's clear that calling it enqueues an email to be sent. You should probably also move the logic in here that checks to make sure the user is in fact confirmed before you fire this off.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,48 @@
...
+        await _applicationContext.SaveChangesAsync(token);
+
+#pragma warning disable CS4014 // Suppress no 'await' call
+        _emailSender.EnqueueEmailAsync(
+            user.Email,
+            // TODO: Finalise the title
+            "Confirmation",
+            GenerateAccountConfirmationEmailBody(user, newConfirmation)
+        );
+#pragma warning restore CS4014
```
TheTedder, line 40: I think you actually need to pay attention to this warning. If you don't use await I'm pretty sure this could run into an issue, especially because you return immediately.

LeaderboardBackend/Services/Impl/UserService.cs
```diff
@@ -23,6 +24,30 @@ public UserService(ApplicationContext applicationContext, IAuthService authServi
         return await _applicationContext.Users.FindAsync(id);
     }
 
+    public async Task<GetUserResult> GetRegisteredUserFromClaims(ClaimsPrincipal claims)
```
TheTedder, line 27: Debating whether we should simply have a GetUserFromClaims method and then just check the returned user's role. That would allow us to avoid duplicating logic.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Moq;
+using NUnit.Framework;
+
+namespace LeaderboardBackend.Test.Features.Users;
+
+[TestFixture]
+public class AccountConfirmationTests : IntegrationTestsBase
+{
+    private const string RESEND_CONFIRMATION_URI = "/account/confirm";
```
TheTedder, line 19: Shouldn't this go in that Consts.cs file we created?

TheTedder (commented):
Few things to look at here. Looking for feedback on a few of the convos I opened.

LeaderboardBackend/Models/Entities/AccountConfirmation.cs
zysim: So I'll set these values on object creation instead? I figured that since these entities are only ever created by the one method in `AccountConfirmationService`, it'd be fine to have these definitions here.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: Yeah, it actually makes more sense for the logic to be in the email method.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: Yeah technically this can throw if the email fails to enqueue, which is something to deal with eventually. Just didn't think it was worth tackling in this PR as well.

zysim: What should the returned error be? 401? 409?

LeaderboardBackend/Services/Impl/UserService.cs
zysim: Yeah; as I've replied in https://github.com/leaderboardsgg/leaderboard-backend/pull/172#discussion_r1296998557, it doesn't really make sense for this method to exist. We can just get the user from claims, and then only verify their role in the create confirmation + send email method.

zysim: I've renamed it to `GetUserFromClaims`

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
TheTedder: If the email fails to enqueue there you could argue for actually using a 500 error since the server itself has encountered an issue. You could also just swallow the error or maybe catch it and trigger a datadog alert or something.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,61 @@
...
+        _applicationContext.AccountConfirmations.Add(newConfirmation);
+
+        await _applicationContext.SaveChangesAsync(token);
+
+        try
+        {
+            await _emailSender.EnqueueEmailAsync(
+                user.Email,
+                // TODO: Finalise the title
+                "Confirmation",
```
TheTedder, line 46: "Confirm Your Account" would be fine.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
TheTedder: You can just leave it as is for now. We can worry about this later.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -152,11 +152,13 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
-            {
-                return Conflict();
-            };
-            return Ok();
+            CreateConfirmationResult r = await confirmationService.CreateConfirmationAndSendEmail(user);
+
+            return r.Match<ActionResult>(
+                confirmation => Ok(),
+                badRole => Conflict(),
+                emailFailed => Conflict()
```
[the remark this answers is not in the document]
TheTedder: If the email fails to send you could return a 500 or you could just hide the error. I would not return 409 here.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: I'll add a todo to handle failed email sends.

TheTedder: Sounds good. Can you make it log an error too?

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,61 @@
...
+        if (user.Role is not UserRole.Registered)
+        {
+            return new BadRole();
+        }
+
+        AccountConfirmation newConfirmation =
+            new()
+            {
+                CreatedAt = SystemClock.Instance.GetCurrentInstant(),
+                ExpiresAt = SystemClock.Instance.GetCurrentInstant() + Duration.FromHours(1),
```
TheTedder, line 33: You could reduce this to one `GetCurrentInstant()` call.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: We don't have a logging library yet. I've put a todo in the service file to handle that when we have a logging library installed.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: Oh yeah lol

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            Email = "email",
+            Password = "password",
+            Username = "username",
+        });
+        string token = _authService.GenerateJSONWebToken(result.AsT0);
+
+        client.DefaultRequestHeaders.Authorization = AuthenticationHeaderValue.Parse($"Bearer {token}");
+        HttpResponseMessage res = await client.PostAsync(Routes.RESEND_CONFIRMATION, null);
+        res.Should().HaveStatusCode(HttpStatusCode.InternalServerError);
+    }
```
TheTedder, line 110: I'm not sure if we need to test something that shouldn't actually happen. We should probably hold off on this test case until we figure out what behavior we actually want in this scenario.

zysim: I'd argue instead that it's worth making sure we have the behaviour we've currently got. When we change the action's behaviour later on, then we change the test.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            x.EnqueueEmailAsync(
+                "email",
+                "Confirm Your Account",
+                It.IsAny<string>()
+            ),
+            Times.Once()
+        );
+        ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+        context.AccountConfirmations.FirstOrDefault(c => c.UserId == result.AsT0.Id)
+            .Should().NotBeNull();
```
Dalet, line 146: Check that the expiration date is 1 hour from now on the created entity.

`IClock` should be injected in `AccountConfirmationService`  instead of using a global instance, to allow `GetCurrentInstant()` to be mocked.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -0,0 +1,148 @@
...
+    public void TearDown()
+    {
+        s_factory.ResetDatabase();
+        _scope.Dispose();
+    }
+
+    [Test]
+    public async Task ResendConfirmation_Unauthorised()
+    {
+
```
Dalet, line 39: Empty line

Dalet (changes requested):
I disagree on some implementations details, but I think we need to move on (and I don't have time to argue lol).
The behavior looks good. Left a comment about expiration date being untested.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
zysim: Can't inject `IClock`. It won't build when I do so. Might need to use the global instance. However there's [NodaTime's testing lib](https://nodatime.org/2.1.x/userguide/testing) that we can use in tests instead.

TheTedder:
> I disagree on some implementations details, but I think we need to move on (and I don't have time to argue lol). The behavior looks good. Left a comment about expiration date being untested.

Write them down for the future and we can come back to it when we have time.

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
Dalet: `IClock` is now injected, but current date is not mocked in the tests to verify the dates?
`CreatedAt` could be `1975/08/22 2PM` and `ExpirationDate` could be `1975/08/22 3 PM` for all we know.

Can use the library's fake implem for testing: https://nodatime.org/1.4.x/api/NodaTime.Testing.FakeClock.html

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,65 @@
...
+            // TODO: Log/otherwise handle the fact that the email failed to be queued - zysim
+            return new EmailFailed();
+        }
+
+        return newConfirmation;
+    }
+
+    // TODO: Finalise message contents
+    private string GenerateAccountConfirmationEmailBody(User user, AccountConfirmation confirmation) =>
+        $@"Hi {user.Username},<br/><br/>Click <a href=""/confirm-account?code={Convert.ToBase64String(confirmation.Id.ToByteArray())}"">here</a> to confirm your account.";
```
TheTedder, line 64: You need to put the full website URL, not just the path; the user will be opening this link from their mail client. Also, use this encoding for GUIDs to ensure url safety and to remove padding: https://datatracker.ietf.org/doc/html/rfc7515#page-55. I suggest adding an extension method for it.

TheTedder (changes requested):

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,65 @@
...
+        _emailSender = emailSender;
+        _clock = clock;
+    }
+
+    public async Task<AccountConfirmation?> GetConfirmationById(Guid id)
+    {
+        return await _applicationContext.AccountConfirmations.FindAsync(id);
+    }
+
+    public async Task<CreateConfirmationResult> CreateConfirmationAndSendEmail(User user, CancellationToken token = default)
```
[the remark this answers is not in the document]
zysim: Figured it might be useful for if `SaveChangesAsync` takes too long, we can cancel the operation from the controller's end.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: I remember reading somewhere in the .NET docs that paths will be relative to the base URL. Of course there's no way to really prove it until we do start sending emails.

I'll switch to the full path, and I'll implement the extension method.

Dalet: > I remember reading somewhere in the .NET docs that paths will be relative to the base URL

Whatever that was, there is no way it could apply to emails.

Gonna need to add required configuration for the frontend base URL. In [AppConfig](https://github.com/leaderboardsgg/leaderboard-backend/blob/main/LeaderboardBackend/AppConfig.cs#L6) seems good.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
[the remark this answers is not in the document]
Dalet: Cancellation tokens passed from the controller are triggered when the client cancels the request (i.e. they close the tab or refresh the page before getting a response). If it's just the `default` token it doesn't do anything.

I prefer using these only for read-only requests or very specific scenarios.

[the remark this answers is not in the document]
zysim: I see. In that case I'll drop the arg.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -59,7 +59,13 @@ public async Task<CreateConfirmationResult> CreateConfirmationAndSendEmail(User
...
 
-    // TODO: Finalise message contents
     private string GenerateAccountConfirmationEmailBody(User user, AccountConfirmation confirmation) =>
-        $@"Hi {user.Username},<br/><br/>Click <a href=""/confirm-account?code={Convert.ToBase64String(confirmation.Id.ToByteArray())}"">here</a> to confirm your account.";
+        $@"Hi {user.Username},<br/><br/>Click <a href=""https://leaderboards.gg/confirm-account?code={EncodeIdForEmail(confirmation)}""here</a> to confirm your account.";
+
+    // Copy of https://datatracker.ietf.org/doc/html/rfc7515#page-55
+    private string EncodeIdForEmail(AccountConfirmation confirmation)
+        => Convert.ToBase64String(confirmation.Id.ToByteArray())
+            .Split('=')[0]
```
TheTedder, line 68: Use `TrimEnd('=')`

LeaderboardBackend/AppConfig.cs
```diff
@@ -4,6 +4,7 @@ namespace LeaderboardBackend;
 
 public class AppConfig
 {
+    public static Uri BASE_PATH = new("https://leaderboards.gg");
```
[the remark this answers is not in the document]
TheTedder: Also, this shouldn't be called BASE_PATH because it's actually the URL of the frontend isn't it?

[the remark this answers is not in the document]
zysim: Yeah it's probably not the best name. Maybe `BASE_URL`?

And I've made it static because I don't see why it should be an instance member, unless injecting AppConfig into the service is preferable.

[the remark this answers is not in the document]
TheTedder: Seeing as it's not the base of _this_ app necessarily, I think WEBSITE_URL or FRONTEND_URL would be preferable.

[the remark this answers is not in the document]
Dalet: I agree with Tedder on naming.

----
Don't make it static, it sucks for testing.
`AppConfig` is already injectable as a `IOptions<AppConfig>`.

Also don't set a default value, it should be set when setting up the app.
A default value should be good for most cases, however here it's only good in the case of a specific environment (lbgg's production).
You can add a `.NotNull()` validation for it.

Do add it to the `example.env` though,  
Setting it to the frontend default local url is nice. `AppConfig__WebsiteUrl=http://localhost:3000` (I don't remember the actual default port) 

----
No need to do that right now, but with this new config we can allow cross origin for this URL instead of relying on the `AllowOrigins` config, and it can thus be made optional.

TheTedder (changes requested):

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        // Copy of https://datatracker.ietf.org/doc/html/rfc7515#page-55
+        string encodedConfirmationId = Convert.ToBase64String(confirmation.Id.ToByteArray())
+            .TrimEnd('=')
+            .Replace('+', '-')
+            .Replace('/', '_');
+        UriBuilder builder = new(_appConfig.WebsiteUrl);
+        builder.Path = "confirm-account";
+        builder.Query = $"code={encodedConfirmationId}";
+        return $@"Hi {user.Username},<br/><br/>Click <a href=""{builder.Uri.ToString()}""here</a> to confirm your account.";
```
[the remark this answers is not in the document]
zysim: 🤦

fly.toml
```diff
@@ -5,6 +5,7 @@ app = "lbgg-backend-dev"
 
 [env]
   AllowedOrigins = "*"
+  WebsiteUrl = "https://leaderboards.gg"
```
Dalet, line 8: Should be netlify here

LeaderboardBackend/appsettings.json
```diff
@@ -6,6 +6,7 @@
 		}
 	},
 	"EnvPath": ".env",
+	"WebsiteUrl": "https://lbgg.netlify.app",
```
Dalet, line 9: There should be no default config for production

Dalet (changes requested):

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,82 @@
...
+            await _emailSender.EnqueueEmailAsync(
+                user.Email,
+                "Confirm Your Account",
+                GenerateAccountConfirmationEmailBody(user, newConfirmation)
+            );
+        }
+        catch
+        {
+            // TODO: Log/otherwise handle the fact that the email failed to be queued - zysim
+            return new EmailFailed();
```
TheTedder, line 64: What's wrong with using the default logging capabilities of ASP.NET? https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging

zysim: Which library should we use? I hesitated on using Console because I thought we wouldn't want it

TheTedder: You could try EventSource.

LeaderboardBackend/appsettings.Staging.json
TheTedder, line 1: I almost feel like this file shouldn't exist and instead the "staging" config should be managed with secrets/env vars on the fly.io end.

Dalet: Out of the scope of the PR, but I agree in principle.

Right now "staging" as an ASP env name is only used by the integration tests. We could move that config into the code.

The fly.io deployment is using only `appsettings.json` (because its ASP env name is "production"), plus secrets and env vars defined in `fly.toml`

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
Dalet: Injected `ILogger<ClassName>` is the way.

But as I said in the email discord forum post, I don't even think this case should be handled at all.

Dalet (approved):

TheTedder (approved):

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -0,0 +1,82 @@
...
+        return newConfirmation;
+    }
+
+    private string GenerateAccountConfirmationEmailBody(User user, AccountConfirmation confirmation)
+    {
+        // Copy of https://datatracker.ietf.org/doc/html/rfc7515#page-55
+        string encodedConfirmationId = Convert.ToBase64String(confirmation.Id.ToByteArray())
+            .TrimEnd('=')
+            .Replace('+', '-')
+            .Replace('/', '_');
```
TheTedder, line 76: Use the new `Guid.ToUrlSafeBase64String()` method that I added in my last PR.

TheTedder (approved):

Dalet (approved):

Issue #159, opened 2023-06-25
Account recovery


Dalet:
```[tasklist]
### Tasks
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/147
- [x] Create the account recovery entity
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/156
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/157
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/158
```

## Model
The account recovery entity represents an auto-generated token a user can use to recover their account.

### Fields
- [x] The account recovery entity must have a non-nullable GUID called `Id` serving as a primary key.
- [x] The account recovery entity must have a non-nullable foreign key called `UserId` referencing the users table.
- [x] The account recovery entity must have a non-nullable auto-generated timestamp called `CreatedAt` that defaults to the current time at the moment of creation.
- [x] The account recovery entity must have a nullable timestamp called `UsedAt` that defaults to NULL.
- [x] The account recovery entity must have a non-nullable timestamp called `ExpiresAt` that defaults to one hour after creation.

AV7708614:
Como puedo hacer un recovery de mi cuenta de Samsung o Gmail si mi número no lo tengo. Creo que mi teléfono esta muy hackeado necesito ayuda y no se a quien Acudir.

PR #173, opened 2023-08-06
Add Account Recovery Model

zysim:
Solves: #159

### DIFF
--- a/LeaderboardBackend/Models/Entities/AccountRecovery.cs
+++ b/LeaderboardBackend/Models/Entities/AccountRecovery.cs
@@ -9,8 +9,6 @@ namespace LeaderboardBackend.Models.Entities;
 /// </summary>
 public class AccountRecovery
 {
-    public static readonly Guid s_seedAdminId = new("421bb896-1990-48c6-8b0c-d69f56d6746a");
-
     /// <summary>
     ///     The unique identifier of the `AccountRecovery`.<br/>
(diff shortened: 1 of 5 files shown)


Dalet (changes requested):

LeaderboardBackend/Models/Entities/AccountRecovery.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public Instant? UsedAt { get; set; }
+
+    /// <summary>
+    ///     The time this `AccountRecovery` expires. Defaults to an hour from its
+    ///     creation.
+    /// </summary>
+    /// <example>john.doe@example.com</example>
+    [Required]
+    public Instant ExpiresAt { get; set; } = SystemClock.Instance.GetCurrentInstant() + Duration.FromHours(1);
```
TheTedder, line 50: This logic probably shouldn't be in the model itself.

zysim:
I'll address the comments once the account confirmation PR's merged.

LeaderboardBackend/Models/Entities/AccountRecovery.cs
zysim: Alright, I'll just remove the inits, and not move them to the service method.

TheTedder (approved):

Dalet (approved):

PR #175, opened 2023-08-11
tests: ensure database initialization before OneTimeSetup of IntegrationTestsBase children runs

Dalet:
Fixes an issue where the testcontainers database is sometimes not yet initialized in OneTimeSetup methods of `IntegrationTestsBase` children. (noticed this issue with #169)

Also upgraded the testcontainers package since a new version has bugfixes for Docker detection. (@zysim had trouble with this)

### DIFF
--- a/LeaderboardBackend.Test/TestApi/TestApiFactory.cs
+++ b/LeaderboardBackend.Test/TestApi/TestApiFactory.cs
@@ -6,7 +6,6 @@ using LeaderboardBackend.Test.Lib;
 using MailKit.Net.Smtp;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Mvc.Testing;
-using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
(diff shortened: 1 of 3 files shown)


LeaderboardBackend.Test/TestApi/TestApiFactory.cs
```diff
@@ -85,6 +69,38 @@ public TestApiClient CreateTestApiClient()
...
+            case DatabaseBackend.TestContainer:
+                if (!PostgresDatabaseFixture.HasCreatedTemplate)
+                {
+                    dbContext.MigrateDatabase();
+                    Seed(dbContext);
+                    PostgresDatabaseFixture.CreateTemplateFromCurrentDb();
+                }
+
+                break;
+            case DatabaseBackend.InMemory:
```
TheTedder, line 92: Are we not dropping support for in-memory DBs due to them not supporting the constraints we need? I seem to remember some discussion around that.

Dalet: We haven't formally decided, though I think we should. But it's not the purpose of the PR.

TheTedder (approved):
I love small PRs.

Issue #153, opened 2023-06-24
Account confirmation


Dalet:
```[tasklist]
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/147
- [x] Create the account confirmation entity
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/154
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/155
- [ ] Send a confirmation email on user registration (#148)
```

## Model
The account confirmation entity represents an auto-generated token a user can use to confirm their account.

### Fields
Fields are non-nullable unless specified.
| Field       | Type               | Description                                        |
|-------------|--------------------|----------------------------------------------------|
| `Id`        | GUID               | Primary key                                        | 
| `UserId`    | GUID               | Id of the user to confirm (FK)                     | 
| `CreatedAt` | timestamp          | Creation date. Auto-generated on creation.         | 
| `ExpiresAt` | timestamp          | Date until when the confirmation link can be used. | 
| `UsedAt`    | nullable timestamp | Date when the confirmation link was used.          |  

## Related issues
- Functional specs: https://github.com/leaderboardsgg/leaderboard-site/issues/556
- Front-end task: https://github.com/leaderboardsgg/leaderboard-site/issues/564

PR #192, opened 2023-09-09
Send Confirmation Email on Register

zysim:
Closes: #153

### DIFF
--- a/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
+++ b/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
@@ -110,7 +110,7 @@ public class SendConfirmationTests : IntegrationTestsBase
         res.Should().HaveStatusCode(HttpStatusCode.InternalServerError);
 
         ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
-        context.AccountConfirmations.Any().Should().Be(false);
+        context.AccountConfirmations.Where(c => c.UserId == result.AsT0.Id).Should().NotBeNull();
     }
 
     [Test]
--- a/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
+++ b/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
@@ -92,6 +92,7 @@ public class SendConfirmationTests : IntegrationTestsBase
             builder.ConfigureTestServices(services =>
             {
                 services.AddScoped(_ => emailSenderMock.Object);
+                services.AddSingleton<IClock, FakeClock>(_ => new(Instant.FromUnixTimeSeconds(1)));
             });
         })
         .CreateClient();
@@ -110,7 +111,11 @@ public class SendConfirmationTests : IntegrationTestsBase
         res.Should().HaveStatusCode(HttpStatusCode.InternalServerError);
 
         ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
-        context.AccountConfirmations.Where(c => c.UserId == result.AsT0.Id).Should().NotBeNull();
+        context.AccountConfirmations.Where(c =>
+            c.UserId == result.AsT0.Id &&
+            c.CreatedAt == Instant.FromUnixTimeSeconds(1) &&
+            c.ExpiresAt == Instant.FromUnixTimeSeconds(1) + Duration.FromHours(1)
+        ).Should().NotBeEmpty();
     }
 
     [Test]
--- a/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
+++ b/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
@@ -108,6 +108,9 @@ public class SendConfirmationTests : IntegrationTestsBase
         client.DefaultRequestHeaders.Authorization = AuthenticationHeaderValue.Parse($"Bearer {token}");
         HttpResponseMessage res = await client.PostAsync(Routes.RESEND_CONFIRMATION, null);
         res.Should().HaveStatusCode(HttpStatusCode.InternalServerError);
+
+        ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+        context.AccountConfirmations.Any().Should().Be(false);
     }
 
     [Test]
--- a/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
+++ b/LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
@@ -92,7 +92,6 @@ public class SendConfirmationTests : IntegrationTestsBase
             builder.ConfigureTestServices(services =>
             {
                 services.AddScoped(_ => emailSenderMock.Object);
-                services.AddSingleton<IClock, FakeClock>(_ => new(Instant.FromUnixTimeSeconds(1)));
             });
         })
         .CreateClient();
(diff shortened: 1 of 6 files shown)


LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -48,18 +49,18 @@ public async Task<CreateConfirmationResult> CreateConfirmationAndSendEmail(User
...
                 GenerateAccountConfirmationEmailBody(user, newConfirmation)
             );
+            await _applicationContext.SaveChangesAsync();
         }
         catch
         {
+            _applicationContext.AccountConfirmations.Entry(newConfirmation).State = EntityState.Detached;
             // TODO: Log/otherwise handle the fact that the email failed to be queued - zysim
             return new EmailFailed();
         }
```
TheTedder, line 66: You're making the assumption here that the exception was thrown when attempting to send the email and not while saving to the database. If you want to ensure that this is the case, catch a specific exception. In fact, I don't think this is going to work at all since `EnqueueEmailAsync` doesn't seem to throw errors.

TheTedder: Here's an excerpt from `EnqueueEmailAsync`. As you can see, any exceptions generated by `SendEmail` are swallowed.

```
try
{
    await SendEmail(message);
}
finally
{
    _semaphore.Release();
}
```

TheTedder: Or maybe not? I guess they're automatically propagated if there's no catch block?

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -48,18 +49,18 @@ public async Task<CreateConfirmationResult> CreateConfirmationAndSendEmail(User
...
             await _emailSender.EnqueueEmailAsync(
                 user.Email,
                 "Confirm Your Account",
                 GenerateAccountConfirmationEmailBody(user, newConfirmation)
             );
+            await _applicationContext.SaveChangesAsync();
         }
         catch
         {
+            _applicationContext.AccountConfirmations.Entry(newConfirmation).State = EntityState.Detached;
```
TheTedder, line 63: I don't think this is necessary. If the call to `SaveChangesAsync` never happens, then the changes should be dropped when the context is disposed ed, which should happen automatically when the thread handling this query terminates.

TheTedder: Alternatively, you can store the return value of calling `Add` in a local var and then use that. Or just call `Clear` on the change tracker.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
Dalet: The previous code was better imo.
As things are right now, it's better to save the confirmation before sending the email. If you send the email and then the DB request fails, the user is getting a broken link in their email.

(once email queueing is implemented, both actions can be in the same transaction and fail together)

TheTedder: I think in that case it's better to just create the confirmation object outside of the try catch then but do it BEFORE. We really don't expect it to fail because if it does then we have bigger issues.

TheTedder: We really don't care about the case when the email fails but the object gets created. You could just not even handle it.

zysim: I figured that we wouldn't want to create a confirmation in the table if the email fails to send (and of course if saving to DB fails, then it wouldn't be added at all). But yeah; forgot about the fact that this change means that an email can be sent even if the DB transaction fails. I'll revert the changes then

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
zysim: See I'm not sure if the query will terminate if I don't change the entity state (by setting it to `Detached` or calling `Clear` on the change tracker). Do we know for sure that it will?

Anyway, as we've discussed in the thread above, I'll just revert this commit entirely.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
TheTedder: If it fails to save changes to the DB, I believe it will through an exception, which will result in the whole thread essentially erroring out and terminating before it even gets to sending the email.

LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
```diff
@@ -108,6 +108,9 @@ public async Task ResendConfirmation_EmailFailedToSend()
         client.DefaultRequestHeaders.Authorization = AuthenticationHeaderValue.Parse($"Bearer {token}");
         HttpResponseMessage res = await client.PostAsync(Routes.RESEND_CONFIRMATION, null);
         res.Should().HaveStatusCode(HttpStatusCode.InternalServerError);
+
+        ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+        context.AccountConfirmations.Where(c => c.UserId == result.AsT0.Id).Should().NotBeNull();
```
TheTedder, line 113: Would it be more technically correct to check for something like Any returning true? I think technically since you're using a Where clause, the result could be empty but not null.

zysim: Just realised that the most thorough thing would be to check for a row that the user ID and timestamps fit the bill.

LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
```diff
@@ -108,6 +109,13 @@ public async Task ResendConfirmation_EmailFailedToSend()
         client.DefaultRequestHeaders.Authorization = AuthenticationHeaderValue.Parse($"Bearer {token}");
         HttpResponseMessage res = await client.PostAsync(Routes.RESEND_CONFIRMATION, null);
         res.Should().HaveStatusCode(HttpStatusCode.InternalServerError);
+
+        ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+        context.AccountConfirmations.Where(c =>
+            c.UserId == result.AsT0.Id &&
+            c.CreatedAt == Instant.FromUnixTimeSeconds(1) &&
+            c.ExpiresAt == Instant.FromUnixTimeSeconds(1) + Duration.FromHours(1)
+        ).Should().NotBeEmpty();
```
TheTedder, line 118: Sorry for being picky but I just realized if the email failed to send then it shouldn't matter if the confirmation was created or not because the user can't use it.

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -48,8 +48,7 @@ public async Task<CreateConfirmationResult> CreateConfirmationAndSendEmail(User
                 UserId = user.Id,
             };
 
-        _applicationContext.AccountConfirmations.Add(newConfirmation);
-
+        await _applicationContext.AccountConfirmations.AddAsync(newConfirmation);
```
TheTedder, line 51: I'm guilty of making this mistake in my own code but plain Add() is actually preferred. See #199.

LeaderboardBackend.Test/Features/Users/SendConfirmationTests.cs
zysim: Yeah, it really doesn't matter to check for it

TheTedder (changes requested):
Add a test or amend an existing test that verifies that the account confirmation was created and the email was sent.

TheTedder (approved):

PR #193, opened 2023-09-12
Confirm Account Endpoint

### LINKED ISSUE #155: Task: Confirm account endpoint
Parent issue: https://github.com/leaderboardsgg/leaderboard-backend/issues/153

Endpoint: `PUT /account/confirm/{confirmationId}`

URL Parameters:
- `confirmationId`: GUID

### Requirements
- [ ] If the confirmation ID provided by the client is not a base64-encoded ID of an existing confirmation record in the database, the server must respond with a `404 NOT FOUND` status code.
- [ ] If the confirmation specified by the client has an `ExpiredAt` timestamp that is in the past, the server must respond with a `404 NOT FOUND` status code
- [ ] If the recovery specified by the client has a non-null `UsedAt` field, the server must respond with a `404 NOT FOUND` status code.
- [ ] If the user associated with the confirmation specified by the client has a role of `CONFIRMED` or `ADMINISTRATOR`, the server must respond with a `409 CONFLICT` status code, indicating that the user account has already been confirmed.
- [ ] If the account is successfully confirmed, the server must set the associated users's role to `CONFIRMED`, set the specified confirmation's `UsedAt` timestamp to the current time, and respond with a `200 OK` status code.

### Procedure
This is what happens when the user confirms their account via a link sent to their email inbox.

Inputs:
- confirmation ID: GUID

- [ ] The user associated with the confirmation identified by the supplied ID must have their `Role` field updated from `REGISTERED` to `CONFIRMED` in the database and the procedure must return `TRUE`.
- [ ] If the user's role was not `REGISTERED`. the user's role must not be altered.

TheTedder:
Closes #155


LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -167,4 +166,23 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
             notFound => Unauthorized()
         );
     }
+
```
Dalet, line 169: Missing the XML comment doc for each status code

Dalet (changes requested):

LeaderboardBackend/Services/Impl/AccountConfirmationService.cs
```diff
@@ -67,6 +68,38 @@ public async Task<CreateConfirmationResult> CreateConfirmationAndSendEmail(User
...
+        }
+
+        if (confirmation.UsedAt is not null)
+        {
+            return new AlreadyUsed();
+        }
+
+        Instant now = _clock.GetCurrentInstant();
+
+        if (confirmation.ExpiresAt < now)
```
Dalet, line 92: I would have expected greater or equal to now since it's "ExpiresAt"

TheTedder: It needs to be less than [or equal to]. Times that are less than other times are in the past, so if the expiration date is less than (in the past) compared to now, then it's expired.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -167,4 +166,32 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+    /// <summary>
+    ///     Confirms a user account.
+    /// </summary>
+    /// <param name="id">The confirmation token.</param>
+    /// <param name="confirmationService">IAccountConfirmationService dependency.</param>
+    /// <response code="200">The account was confirmed successfully.</response>
+    /// <response code="400">The token was malformed.</response>
```
zysim, line 176: Our Guid model binder should trigger this, right?

TheTedder: I believe so but I can put an additional test in if you want.

zysim: Yes, please. We don't have a test that checks for this currently as well.

TheTedder: Actually, 404 would be more appropriate here, since the resource we're trying to locate could not be found. From the client's perspective, the format of the token is an implementation detail and does not matter.

zysim: In that case, you could remove that line then

zysim (dismissed):

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -151,4 +153,191 @@ public async Task ResendConfirmation_Success()
...
+        };
+
+        ApplicationContext context = _scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+        await context.AccountConfirmations.AddAsync(confirmation);
+        await context.SaveChangesAsync();
+        HttpResponseMessage res = await client.PutAsync(Routes.ConfirmAccount(confirmation.Id), null);
+        res.Should().HaveStatusCode(HttpStatusCode.OK);
+        context.ChangeTracker.Clear();
+        AccountConfirmation? conf = await context.AccountConfirmations.Include(c => c.User).SingleOrDefaultAsync(c => c.Id == confirmation.Id);
+        conf!.UsedAt.Should().NotBeNull();
```
Dalet, line 340: Should check that `UsedAt` is equal to `Instant.FromUnixTimeSeconds(1)`

LeaderboardBackend.Test/Features/Users/AccountConfirmationTests.cs
```diff
@@ -151,4 +153,191 @@ public async Task ResendConfirmation_Success()
         confirmation.UsedAt.Should().BeNull();
         Instant.Subtract(confirmation.ExpiresAt, confirmation.CreatedAt).Should().Be(Duration.FromHours(1));
     }
+
```
Dalet, line 156: Need to split these tests to a new file.

Once that's done, the `HttpClient` creation can be written only once in the file's `[SetUp]` method.
You can store the `FakeClock` instance in a field and use it in the tests to change the time when you need to (`clock.Reset(instant)`).

Dalet (commented):
Left some non-critical comments, feel free to merge without addressing them.

Dalet (approved):

PR #195, opened 2023-09-13
Account Recovery Endpoint

### LINKED ISSUE #156: Task: Send recovery email endpoint

```[tasklist]
### Blocked by
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/147
```

Endpoint: `POST /account/recover`

The server may impose a limit on how often a client can call this endpoint. If the client calls this endpoint too many times within a given timespan, the server may respond with a `429 TOO MANY REQUESTS` status code.

Request body fields:
* email: string
* username: string

- [ ] The server must respond with a `200 OK` status code, even if the specified username and email do not correspond to an existing user account. This is to prevent malicious users from determining another user's email and username.
- [ ] If the email and username specified belong to the same existent user whose role is `CONFIRMED` or `ADMIN`, the server must initiate the account recovery link generation procedure. (See below.)

## Procedure
Inputs:
* email: string
* username: string

- [ ] If the email and username supplied to this procedure do not both correspond to the same existent user in the database, the procedure must terminate without performing any further actions.
- [ ] If the corresponding user has a role of `BANNED` or `REGISTERED`, the procedure must terminate without performing any further actions.
- [ ] A new recovery record must be created in the database.
  - [ ] The new recovery record's `UserId` field must be the ID corresponding to the user with the specified email and username.
- [ ] An email must be sent to the supplied email address,
  -  TODO: Determine the exact contents of the email.
  - [ ] The email must contain a link to the password reset page.
    - [ ] The password reset page link must have the path `/reset-password?code={id}` where `{id}` is the newly-created recovery's ID in Base64 encoding.

TheTedder:
Closes #156

### DIFF
--- a/LeaderboardBackend/Controllers/AccountController.cs
+++ b/LeaderboardBackend/Controllers/AccountController.cs
@@ -170,4 +170,28 @@ public class AccountController : ApiController
             notFound => Unauthorized()
         );
     }
+
+    /// <summary>
+    ///     Sends an account recovery email.
+    /// </summary>
+    /// <param name="recoveryService">IAccountRecoveryService dependency.</param>
+    /// <param name="request">The account recovery request.</param>
+    /// <response code="200">This endpoint returns 200 OK regardless of whether the email was sent successfully or not.</response>
+    [AllowAnonymous]
+    [HttpPost("recover")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    public async Task<ActionResult> RecoverAccount(
+        [FromServices] IAccountRecoveryService recoveryService,
+        [FromBody] RecoverAccountRequest request
+    )
+    {
+        User? user = await _userService.GetUserByNameAndEmail(request.Username, request.Email);
+
+        if (user is not null)
+        {
+            await recoveryService.CreateRecoveryAndSendEmail(user);
+        }
+
+        return Ok();
+    }
 }
--- a/LeaderboardBackend/Controllers/AccountController.cs
+++ b/LeaderboardBackend/Controllers/AccountController.cs
@@ -2,6 +2,7 @@ using LeaderboardBackend.Controllers.Annotations;
 using LeaderboardBackend.Models.Entities;
 using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Models.ViewModels;
+using LeaderboardBackend.Result;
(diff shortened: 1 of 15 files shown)


LeaderboardBackend/Services/Impl/AccountRecoveryService.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        catch
+        {
+            return new EmailFailed();
+        }
+
+        return recovery;
+    }
+
+    private string GenerateAccountRecoveryEamilBody(User user, AccountRecovery recovery)
```
zysim, line 63: Typo
```suggestion
    private string GenerateAccountRecoveryEmailBody(User user, AccountRecovery recovery)
```

LeaderboardBackend/openapi.json
```diff
@@ -193,6 +193,49 @@
...
+            "text/json": {
+              "schema": {
+                "$ref": "#/components/schemas/RecoverAccountRequest"
+              }
+            },
+            "application/*+json": {
+              "schema": {
+                "$ref": "#/components/schemas/RecoverAccountRequest"
+              }
+            }
```
zysim, line 219: Don't think these should be generated.

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -169,4 +170,30 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+    )
+    {
+        User? user = await _userService.GetUserByNameAndEmail(request.Username, request.Email);
+
+        if (user is not null)
+        {
+            await recoveryService.CreateRecoveryAndSendEmail(user);
+        }
+
+        return Ok();
```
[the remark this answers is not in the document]
TheTedder: Don't want people to be able to find other users' email addresses this way. We haven't worried about this too much with other endpoints but we will need to.

LeaderboardBackend/openapi.json
TheTedder: I don't understand. The endpoint takes a JSON payload in the body.

TheTedder: I fixed it anyway.

zysim (approved):

PR #196, opened 2023-10-08
Test Recovery Code Endpoint

### LINKED ISSUE #157: Task: Test Password Reset Link endpoint

`GET /account/recover/{id}`

URL Parameters:
* id: GUID

- [ ] If the ID provided by the client is not a base64-encoded ID of an existing recovery record in the database, the server must respond with a `404 NOT FOUND` status code.
- [ ] If the recovery specified by the client has an `ExpiresAt` timestamp that is in the past, the server must respond with a `404 NOT FOUND` status code. (Gone would probably work fine here too.)
- [ ] If the recovery specified by the client is not the most recently-created recovery, the server must respond with a `404 NOT FOUND` status code.
- [ ] If the recovery specified by the client has a non-null `UsedAt` field, the server must respond with a `404 NOT FOUND` status code.
- [ ] If the user associated with the recovery specified by the client has a role of `BANNED`, the server must respond with a `404 NOT FOUND` status code.
- [ ] If none of the above error conditions are met, the server must respond with a `200 OK` status code.

TheTedder:
Closes #157

### DIFF
--- a/LeaderboardBackend.Test/Features/Users/RecoverAccountTests.cs
+++ b/LeaderboardBackend.Test/Features/Users/RecoverAccountTests.cs
@@ -1,0 +1,178 @@
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using LeaderboardBackend.Models.Entities;
+using LeaderboardBackend.Test.Fixtures;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using NodaTime;
+using NodaTime.Testing;
+using NUnit.Framework;
+
+namespace LeaderboardBackend.Test.Features.Users;
+
+public class RecoverAccountTests : IntegrationTestsBase
+{
+    private IServiceScope _scope = null!;
+    private HttpClient _client = null!;
+    private readonly FakeClock _clock = new(Instant.FromUnixTimeSeconds(1));
+
+    [SetUp]
+    public void Init()
+    {
+        _scope = _factory.Services.CreateScope();
+
+        _client = _factory.WithWebHostBuilder(builder =>
+        {
+            builder.ConfigureTestServices(services =>
+            {
+                services.AddSingleton<IClock, FakeClock>(_ => _clock);
+            });
+        }).CreateClient();
+    }
+
+    [TearDown]
+    public void TearDown()
+    {
+        _factory.ResetDatabase();
+        _scope.Dispose();
+    }
+
+    [TestCase("not_a_guid")]
+    [TestCase("L8msfy9wd0qWbDJMZwwgQg")]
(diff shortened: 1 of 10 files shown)


LeaderboardBackend/Services/Impl/AccountRecoveryService.cs
```diff
@@ -70,4 +72,46 @@ private string GenerateAccountRecoveryEmailBody(User user, AccountRecovery recov
...
+        AccountRecovery? recovery = await _applicationContext.AccountRecoveries.Include(ar => ar.User).SingleOrDefaultAsync(ar => ar.Id == id);
+
+        if (recovery is null)
+        {
+            return new NotFound();
+        }
+
+        if (recovery.User.Role is not UserRole.Registered)
+        {
+            return new BadRole();
```
Dalet, line 87: According to the spec, only the `Banned` role should prevent account recovery. But you probably would have noticed this if the tests were actually running :)

Dalet (changes requested):

LeaderboardBackend.Test/Features/Users/RecoverAccountTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                Username = "username",
+                Role = UserRole.Confirmed
+            }
+        };
+
+        await context.AccountRecoveries.AddAsync(recovery);
+        await context.SaveChangesAsync();
+        HttpResponseMessage res = await _client.GetAsync(Routes.RecoverAccount(recovery.Id));
+        res.Should().HaveStatusCode(HttpStatusCode.OK);
+    }
```
Dalet, line 177: Might wanna assert that the recovery entity was not modified.

Easily achievable with `newlyQueriedRecovery.Should().BeEquivalentTo(originalRecovery)`

TheTedder: I mean shouldn't that be true for not just every test case but every GET request? In general, an HTTP GET request should never modify state.

LeaderboardBackend.Test/Features/Users/TestRecoveryTests.cs
```diff
@@ -0,0 +1,160 @@
...
+
+public class TestRecoveryTests : IntegrationTestsBase
+{
+    private IServiceScope _scope = null!;
+    private readonly HttpClient _client;
+    private readonly FakeClock _clock = new(Instant.FromUnixTimeSeconds(1));
+
+    public TestRecoveryTests()
+    {
+        _client = _factory.WithWebHostBuilder(builder =>
```
Dalet, line 22: Avoid using the constructor with NUnit and use a [OneTimeSetUp] method instead.
As opposed to XUnit, NUnit doesn't guarantee that it will run the constructor only once.

Dalet (approved):

PR #198, opened 2023-10-17
Password Reset Endpoint

### LINKED ISSUE #158: Task: Reset password endpoint

`POST /account/recover/{id}`

URL Parameters:
* id: GUID

Request body fields:
* password: string

- [x] If the ID provided by the client is not a base64-encoded ID of an existing recovery record in the database, the server must respond with a `404 NOT FOUND` status code.
- [x] If the recovery specified by the client has an `ExpiresAt` timestamp that is in the past, the server must respond with a `404 NOT FOUND` status code. (Gone would probably work fine here too.)
- [x] If the recovery specified by the client is not the most recently-created recovery, the server must respond with a `404 NOT FOUND` status code.
- [x] If the recovery specified by the client has a non-null `UsedAt` timestamp, the server must respond with a `404 NOT FOUND` status code.
- [x] If the user associated with the recovery specified by the client has a role of `BANNED`, the server must respond with a `403 FORBIDDEN` status code.
- [x] If the password provided by the client is invalid, the server must respond with a `422 UNPROCESSABLE CONTENT` status code. See #148 for a description of a valid password.
- [x] If the password provided by the client is the same as the password of the associated user, the server must respond with a `409 CONFLICT` status code.
- [x] If none of the above error conditions are met, the server must update the associated user's password, set the `UsedAt` field of the specified recovery to the current time, and return a `200 OK` status code.

TheTedder:
Closes #158

### DIFF
--- a/LeaderboardBackend.Test/Features/Users/ResetPasswordTests.cs
+++ b/LeaderboardBackend.Test/Features/Users/ResetPasswordTests.cs
@@ -248,12 +248,12 @@ public class ResetPasswordTests : IntegrationTestsBase
         res.Should().HaveStatusCode(System.Net.HttpStatusCode.UnprocessableEntity);
         ValidationProblemDetails? content = await res.Content.ReadFromJsonAsync<ValidationProblemDetails>();
         content.Should().NotBeNull();
-        
+
         content!.Errors.Should().BeEquivalentTo(new Dictionary<string, string[]>
         {
             { nameof(RegisterRequest.Password), new[] { "PasswordFormat" } }
         });
-        
+
         context.ChangeTracker.Clear();
         recovery = await context.AccountRecoveries.Include(ar => ar.User).SingleAsync(ar => ar.Id == recovery.Id);
         recovery.UsedAt.Should().BeNull();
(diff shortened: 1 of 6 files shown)


LeaderboardBackend/Models/Requests/UserRequests.cs
```diff
@@ -116,6 +116,12 @@ public record RecoverAccountRequest
     public required string Email { get; set; }
 }
 
+public record ChangePasswordRequest
+{
+    [Required]
```
Dalet, line 121: DataAnnotations attributes are ignored, only FluentValidation validators are applied, better remove it to avoid confusion

LeaderboardBackend.Test/Features/Users/ResetPasswordTests.cs
```diff
@@ -0,0 +1,321 @@
...
+
+        context.AccountRecoveries.Add(recovery);
+        await context.SaveChangesAsync();
+
+        HttpResponseMessage res = await _client.PostAsJsonAsync(Routes.RecoverAccount(recovery.Id), new ChangePasswordRequest
+        {
+            Password = pwd
+        });
+
+        res.Should().HaveStatusCode(System.Net.HttpStatusCode.UnprocessableEntity);
```
Dalet, line 246: Must also assert that the `PasswordFormat` error code is returned.

You can pretty much copy/paste from this existing registration test: https://github.com/leaderboardsgg/leaderboard-backend/blob/main/LeaderboardBackend.Test/Features/Users/RegistrationTests.cs#L94-L99

Dalet (changes requested):
An assertion in a test should be added, otherwise it looks good.

PR #203, opened 2023-11-07
Swashbuckle annotations

### LINKED ISSUE #168: Use Swashbuckle.AspNetCore.Annotations
Turns out, Swashbuckle has an [annotations package](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/README.md#swashbuckleaspnetcoreannotations). We can use it to replace the existing XML comments in controllers with easier-to-read annotations, to generate our Open API doc.

### Affected Functionality

- The generated Open API doc

### Other Relevant Issues

N/A

TheTedder:
Closes #168


zysim:
I honestly don't know if I prefer the look of this change. I feel bad that you've already put in the work on this, but things kinda feels less readable, actually.

TheTedder:
> I honestly don't know if I prefer the look of this change. I feel bad that you've already put in the work on this, but things feel less readable, personally. On top of that, the attributes can't be folded away, which gets distracting.

I updated it just in case you change your mind. I think the openapi doc is actually more correct this way.

zysim:
Honestly; this is fine. I'll merge this.

zysim (approved):

zysim (approved):
Good enough for me. I like that we don't need to needlessly-annotate injected services anymore, as well. I wish we could make the error response schema more accurate (e.g. they all show `status: 0` instead of the correct error code), but it's whatever. You can always just test out the API itself in the Swagger doc, anyway.

PR #206, opened 2023-12-14
Add Debug Logs on Account Recovery

zysim:
Adds:
* A little bit of logs to `AccountController` and `AccountRecoveryService`

### DIFF
--- a/LeaderboardBackend/Controllers/AccountController.cs
+++ b/LeaderboardBackend/Controllers/AccountController.cs
@@ -209,10 +209,11 @@ public class AccountController : ApiController
 
         if (user is null)
         {
-            logger.LogError("User not found: {usermame}", request.Username);
+            logger.LogWarning("Account recovery attempt failed. User not found: {username}", request.Username);
         }
         else
         {
+            logger.LogInformation("Sending account recovery email to user: {id}", user.Id);
             await recoveryService.CreateRecoveryAndSendEmail(user);
         }
 
--- a/LeaderboardBackend/Controllers/AccountController.cs
+++ b/LeaderboardBackend/Controllers/AccountController.cs
@@ -190,6 +190,7 @@ public class AccountController : ApiController
     ///     Sends an account recovery email.
     /// </summary>
     /// <param name="recoveryService">IAccountRecoveryService dependency.</param>
+    /// <param name="logger"></param>
     /// <param name="request">The account recovery request.</param>
     /// <response code="200">This endpoint returns 200 OK regardless of whether the email was sent successfully or not.</response>
     /// <response code="400">The request object was malformed.</response>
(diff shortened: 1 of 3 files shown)


TheTedder:
Partially censoring personally identifying info is not acceptable. It needs to be either fully redacted or not present. User names and IDs are not personally identifying info, so we can log those safely instead of emails.

LeaderboardBackend/Services/Impl/AccountRecoveryService.cs
```diff
@@ -55,8 +59,9 @@ public async Task<CreateRecoveryResult> CreateRecoveryAndSendEmail(User user)
                 GenerateAccountRecoveryEmailBody(user, recovery)
             );
         }
-        catch
+        catch (Exception e)
         {
+            _logger.LogError("Recovery email failed to send: {err}", e.Message);
```
TheTedder, line 64: I would be wary of logging the contents of the error message because it could contain the recipient email address or the contents of the email. Instead, consider logging the user ID and optionally the username so we know who was trying to recover their account.

TheTedder (changes requested):

LeaderboardBackend/Controllers/AccountController.cs
```diff
@@ -200,12 +201,17 @@ public async Task<ActionResult<LoginResponse>> Login([FromBody] LoginRequest req
...
+        [FromServices] ILogger<AccountController> logger,
         [FromBody] RecoverAccountRequest request
     )
     {
         User? user = await _userService.GetUserByNameAndEmail(request.Username, request.Email);
 
-        if (user is not null)
+        if (user is null)
+        {
+            logger.LogError("User not found: {usermame}", request.Username);
```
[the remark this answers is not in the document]
TheTedder: Also, `usermame`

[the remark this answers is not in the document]
zysim: You can't log the ID in that line, since there's no user in the first place. Unless you mean in the context of logging the successful case.

Also yeah I could switch to a warning instead of an error

LeaderboardBackend/Services/Impl/AccountRecoveryService.cs
zysim: While that's true, I don't think just knowing who it is would help diagnosing much. We need some way to fish the problem out better. The underlying service call can fail for a lot of reasons, including failure to connect to the SMTP server, or to log into it.

I'll change to logging the exception type, user ID, and name

LeaderboardBackend/Controllers/AccountController.cs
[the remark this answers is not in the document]
TheTedder: You're right. I was mistaken on that one. Warning or info is more appropriate.

LeaderboardBackend/Services/Impl/AccountRecoveryService.cs
```diff
@@ -55,8 +59,14 @@ public async Task<CreateRecoveryResult> CreateRecoveryAndSendEmail(User user)
                 GenerateAccountRecoveryEmailBody(user, recovery)
             );
         }
-        catch
+        catch (Exception e)
         {
+            _logger.LogError(
+                "Recovery email failed to send with {type} for user {id}, {username}",
```
TheTedder, line 65: Could you alter it slightly to make it clear that first thing that's being logged is the type of exception? Something like "Encountered error {type}" or "threw {type}" would work.

TheTedder (changes requested):

LeaderboardBackend/Services/Impl/AccountRecoveryService.cs
zysim: Actually just {type} at the start should do.

TheTedder (approved):

PR #213, opened 2024-06-24
.NET 8

TheTedder:
I took the liberty of upgrading to .NET 8 since .NET 7 is out of support. I also updated almost all of our dependencies. This is bound to break something so it's a good idea to test this heavily. The default placeholder JWT private key had to be increased in length because it didn't have enough entropy or something.

### DIFF
--- a/LeaderboardBackend/Models/Entities/ApplicationContext.cs
+++ b/LeaderboardBackend/Models/Entities/ApplicationContext.cs
@@ -1,3 +1,4 @@
+using System.Data;
 using System.Reflection;
 using Microsoft.EntityFrameworkCore;
 using Npgsql;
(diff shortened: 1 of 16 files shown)


zysim:
1. Gotta allow anonymous fetching of Categories. We really don't need to be authorised for that
2. How do you generate `openapi.json` again? It's been so long

LeaderboardBackend/Models/Entities/ApplicationContext.cs
```diff
@@ -31,9 +32,22 @@ public ApplicationContext(DbContextOptions<ApplicationContext> options)
     public void MigrateDatabase()
     {
         Database.Migrate();
+        bool tempConnection = false;
+        NpgsqlConnection connection = (NpgsqlConnection)Database.GetDbConnection();
```
zysim, line 36: When does the connection drop?

TheTedder: I don't understand the question. Can you rephrase?

zysim:
Will do more testing tomorrow

TheTedder:
> 1. Gotta allow anonymous fetching of Categories. We really don't need to be authorised for that
> 
>     2. How do you generate `openapi.json` again? It's been so long
> 
> 
> Also I've done some testing of the API. Not like "running tests" testing, but calling the API itself, locally. Able to register a new user, log in/out, and create leaderboards and categories, but only if we're authorised. So far so good.
> 
> All tests pass, too.

This PR doesn't touch category logic. Also, we're gonna go through and change all that anyway after account-related flows are done.

zysim:
> > 1. Gotta allow anonymous fetching of Categories. We really don't need to be authorised for that
> 
> This PR doesn't touch category logic. Also, we're gonna go through and change all that anyway after account-related flows are done.

Yeah that's fair. I'll add a ticket for that if there isn't one, already.

zysim:
Yeah the backend works. Only problems are some action behaviours. I'll list them out in a separate ticket.

zysim (approved):

PR #215, opened 2024-07-10
Fix Anon Endpoints

TheTedder:
This PR makes endpoints that were erroneously requiring authorization work properly.

### DIFF
--- a/LeaderboardBackend/Authorization/MiddlewareResultHandler.cs
+++ b/LeaderboardBackend/Authorization/MiddlewareResultHandler.cs
@@ -15,9 +15,9 @@ public class MiddlewareResultHandler : IAuthorizationMiddlewareResultHandler
         PolicyAuthorizationResult policyAuthorizationResult
     )
     {
-        if (policyAuthorizationResult.Forbidden)
+        if (policyAuthorizationResult.AuthorizationFailure?.FailureReasons.Any(fr => fr.Handler is UserTypeAuthorizationHandler) ?? false)
         {
-            httpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+            httpContext.Response.StatusCode = (int)HttpStatusCode.Unauthorized;
             return;
         }
 
--- a/LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
+++ b/LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
@@ -1,5 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
 using LeaderboardBackend.Models.Entities;
+using LeaderboardBackend.Result;
 using LeaderboardBackend.Services;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.Extensions.Options;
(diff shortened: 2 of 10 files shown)


zysim (commented):
Just a bunch of stuff. Also; do endpoints allow anon by default now?

LeaderboardBackend/Program.cs
```diff
@@ -221,40 +221,26 @@
...
             policy.AuthenticationSchemes.Add(JwtBearerDefaults.AuthenticationScheme);
             policy.RequireAuthenticatedUser();
             policy.Requirements.Add(new UserTypeRequirement(UserTypes.ADMINISTRATOR));
         }
-    );
-    options.AddPolicy(
-        UserTypes.MODERATOR,
-        policy =>
+)
+    .AddPolicy(UserTypes.MODERATOR, policy =>
```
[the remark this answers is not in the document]
TheTedder: I don't want to make any unnecessary changes right now. I just want to fix the issue.

LeaderboardBackend/Authorization/MiddlewareResultHandler.cs
```diff
@@ -15,9 +15,9 @@ public class MiddlewareResultHandler : IAuthorizationMiddlewareResultHandler
         PolicyAuthorizationResult policyAuthorizationResult
     )
     {
-        if (policyAuthorizationResult.Forbidden)
+        if (policyAuthorizationResult.AuthorizationFailure?.FailureReasons.Any(fr => fr.Handler is UserTypeAuthorizationHandler) ?? false)
         {
-            httpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+            httpContext.Response.StatusCode = (int)HttpStatusCode.Unauthorized;
```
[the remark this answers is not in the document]
TheTedder: I have no idea at this point. Everything about this auth system feels like a hack. The way I have it now seems to make sense. If it can't validate your JWT, it returns 401 because the server doesn't know who you are.

TheTedder:
> Just a bunch of stuff. Also; do endpoints allow anon by default now?

I believe they do, which makes sense because most of our endpoints are just fetching public data anyway.

LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
```diff
@@ -34,37 +26,27 @@ UserTypeRequirement requirement
...
-        return;
+            if (Handle(user, requirement))
+            {
+                context.Succeed(requirement);
+            }
+            else
+            {
+                context.Fail();
+            }
+        }, badCredentials => context.Fail(new(this, "Bad Credentials")), notFound => context.Fail(new(this, "User Not Found")));
```
zysim, line 41: So if I understand this correctly; in all three failure cases, we will ultimately return a 401 to the client, right?

TheTedder: Not exactly. In the case where the user simply does not have permission to access the resource, 403 will be returned. Not much point in hiding endpoints with 404 since the code is open source and everyone can see which endpoints exist anyway.

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -39,7 +39,7 @@ public async Task<ActionResult<CategoryViewModel>> GetCategory(long id)
     [HttpPost]
```
zysim, line 39: Would need annotating here, too

LeaderboardBackend/Authorization/UserTypeAuthorizationHandler.cs
```diff
@@ -34,10 +34,6 @@ UserTypeRequirement requirement
             {
                 context.Succeed(requirement);
             }
-            else
-            {
-                context.Fail();
-            }
```
zysim: Why don't we want to explicitly fail here? Doesn't seem to affect auth behaviour.

TheTedder: You don't need to call fail.

zysim (approved):

PR #216, opened 2024-07-18
Fix NodaTime Serialization

TheTedder:
This PR fixes an issue where NodaTime types such as `Instant` and `Duration` were being serialized as objects instead of string values. It also gets rid of the schema filter that marks all non-nullable types as required in the openapi spec. It's better to just use the `required` keyword so there's no ambiguity.


LeaderboardBackend/Program.cs
```diff
@@ -189,8 +194,11 @@
     );
 
     c.SupportNonNullableReferenceTypes();
-    c.SchemaFilter<RequiredNotNullableSchemaFilter>();
     c.MapType<Guid>(() => new OpenApiSchema { Type = "string", Pattern = "^[a-zA-Z0-9-_]{22}$" });
+    c.ConfigureForNodaTimeWithSystemTextJson(jsonSerializerOptions, null, null, true, new(DateTimeZoneProviders.Tzdb)
+    {
+        Instant = Instant.FromUtc(1984, 1, 1, 0, 0),
```
zysim, line 200: Why this date, specifically?

TheTedder: There just needs to be a date to use in the "example" field of the schema, otherwise it uses the current date, which breaks the CI. So I chose literally 1984.

zysim: I cannot believe that flew over my head

Issue #180, opened 2023-08-11
Update Leaderboard Model


TheTedder:
The leaderboard model needs to be brought in line with the [DB design](https://lucid.app/documents/view/b2eeb6e7-b1e5-4e10-bfbb-84e968db0fa0).

- [ ] Rename the Rules field to Info.
- [ ] Create the CreatedAt field.
- [ ] Create the UpdatedAt field.
- [ ] Create the DeletedAt field.

Issue #179, opened 2023-08-11
Update Category Model


TheTedder:
The category model needs to be brought in line with the [DB design](https://lucid.app/documents/view/b2eeb6e7-b1e5-4e10-bfbb-84e968db0fa0)

- [ ] Rename the Rules field to Info.
- [ ] Delete the PlayersMin field.
- [ ] Delete the PlayersMax field.
- [ ] Create the Sort Dir field. Use an enum.
- [ ] Create the Type field. Use an enum.
- [ ] Create the CreatedAt field.
- [ ] Create the UpdatedAt field.
- [ ] Create the DeletedAt field.

PR #217, opened 2024-07-22
Update Models

### LINKED ISSUE #39: Run Model
Bring the run entity in line with [DB design](https://lucid.app/lucidchart/b2eeb6e7-b1e5-4e10-bfbb-84e968db0fa0/edit?page=0_0&invitationId=inv_8fd08c8f-0f40-43aa-ac55-91eea160ca6d#).
- [ ] Add TimeOrScore field.
- [ ] Add Info field.
- [ ] Rename SubmittedAt field to CreatedAt.
- [ ] Add UpdatedAt field.
- [ ] Add DeletedAt field.

There may be some infrastructure needed so that soft deletes work properly.

TheTedder:
This PR updates all the models in the database so that they're in line with the [design document](https://lucid.app/lucidchart/b2eeb6e7-b1e5-4e10-bfbb-84e968db0fa0/edit?page=0_0&invitationId=inv_8fd08c8f-0f40-43aa-ac55-91eea160ca6d#). Closes #39, #179, and #180.


LeaderboardBackend/Migrations/20240722190611_RunAddUserId.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<Guid>(
+                name: "user_id",
+                table: "runs",
+                type: "uuid",
+                nullable: false,
+                defaultValue: new Guid("00000000-0000-0000-0000-000000000000"));
```
Eein, line 19: Should this provide a default value when its not nullable?

I can't remember the specific order of operations, but this may return an fkey error instead of a null value check error when interacting with the database.

If the intention is to allow records without a user, just make it nullable (imo)

LeaderboardBackend.Test/Categories.cs
```diff
@@ -68,9 +71,12 @@ public static async Task CreateCategory_GetCategory_OK()
             {
                 Body = new CreateCategoryRequest()
                 {
-                    Name = Generators.GenerateRandomString(),
-                    Slug = Generators.GenerateRandomString(),
-                    LeaderboardId = createdLeaderboard.Id
+                    Name = "1 Player",
+                    Slug = "1_player",
```
Eein, line 75: Some additional test cases referenced here: https://github.com/leaderboardsgg/leaderboard-backend/issues/219

Better to do as isolated unit of work.

LeaderboardBackend/Migrations/20240718220723_CategoryAddTimestamps.cs
```diff
@@ -0,0 +1,50 @@
...
+
+#nullable disable
+
+namespace LeaderboardBackend.Migrations
+{
+    /// <inheritdoc />
+    public partial class CategoryAddTimestamps : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
```
Eein, line 12: In the future when data needs to persist, if EntityFramework has the ability to update column types, postgres is extremely good at coercing types for you in the majority of cases, so you don't need to drop tables. (this is one instance)

Eein (approved):
Added a few notes

LeaderboardBackend/Migrations/20240722190611_RunAddUserId.cs
TheTedder: That's auto-generated. I don't think postgres allows not having any default value. It'll still violate the FK constraint so I'm not worried.

LeaderboardBackend/Migrations/20240718220723_CategoryAddTimestamps.cs
TheTedder: These migrations are autogenerated. I did make sure to look at them though to make sure they were coercing where necessary.

Eein: Be wary of a table drop for production code once things start populating -

Also a note to add db backups for prod :)

TheTedder: We're not dropping here. This is adding columns. Dropping only occurs if you roll this migration back.

Eein: OH

These are new columns - ignore me, and apologies. 

I'm not sure why i thought these had already existed. Sorry!

LeaderboardBackend/Migrations/20240722190611_RunAddUserId.cs
zysim: I'd imagine we'll set the relation to null if the user's been deleted, right? We need to factor that in. Unless we delete the run, as well

TheTedder: There's no current plan for actually deleting user records. We can just ban them. Even if there was, we could just use soft deletes like for lbs, cats, and runs.

LeaderboardBackend/Migrations/20240718212110_AddLeaderboardTimestamps.cs
```diff
@@ -0,0 +1,50 @@
...
+            migrationBuilder.AddColumn<Instant>(
+                name: "deleted_at",
+                table: "leaderboards",
+                type: "timestamp with time zone",
+                nullable: true);
+
+            migrationBuilder.AddColumn<Instant>(
+                name: "updated_at",
+                table: "leaderboards",
+                type: "timestamp with time zone",
```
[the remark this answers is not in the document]
TheTedder: `created_at` is non-nullable so it needs to have a default. These default values are kind of "dumb" right now because I haven't written the code that would install a trigger to properly initialize them yet. That's coming in a later PR because this one was already so bloated that I didn't want to add any more code to it.

LeaderboardBackend/Migrations/20240718222616_CategoryAddSortDirection.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterDatabase()
+                .Annotation("Npgsql:CollationDefinition:case_insensitive", "und-u-ks-level2,und-u-ks-level2,icu,False")
+                .Annotation("Npgsql:Enum:sort_direction", "ascending,descending")
+                .Annotation("Npgsql:Enum:user_role", "registered,confirmed,administrator,banned")
+                .OldAnnotation("Npgsql:CollationDefinition:case_insensitive", "und-u-ks-level2,und-u-ks-level2,icu,False")
+                .OldAnnotation("Npgsql:Enum:user_role", "registered,confirmed,administrator,banned");
```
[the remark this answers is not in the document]
TheTedder: I have no idea. These migrations are autogenerated. I wouldn't worry about it.

LeaderboardBackend/Migrations/20240722190611_RunAddUserId.cs
```diff
@@ -0,0 +1,51 @@
...
+                table: "runs",
+                column: "user_id");
+
+            migrationBuilder.AddForeignKey(
+                name: "fk_runs_users_user_id",
+                table: "runs",
+                column: "user_id",
+                principalTable: "users",
+                principalColumn: "id",
+                onDelete: ReferentialAction.Cascade);
```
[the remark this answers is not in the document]
TheTedder: Users never get deleted so we don't have to worry about it.

LeaderboardBackend/Models/Entities/Run.cs
```diff
@@ -14,26 +13,54 @@ public class Run
...
     public Guid Id { get; set; }
 
+    /// <summary>
+    ///     User-provided details about the run.
+    /// </summary>
+    public string? Info { get; set; }
+
+    public RunType Type => Category.Type;
+
+    public Duration Time => Duration.FromNanoseconds(TimeOrScore);
```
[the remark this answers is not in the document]
TheTedder: It's easier to just do it like this. It's a computed property that you only access if it makes sense to. Kind of like a C union. Best not to over-engineer it.

LeaderboardBackend/Models/Validation/SlugRule.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public const string SLUG_FORMAT = "SlugFormat";
+
+    public static IRuleBuilderOptions<T, string> Slug<T>(this IRuleBuilder<T, string> ruleBuilder)
+        => ruleBuilder
+            .Length(2, 80)
+                .WithErrorCode(SLUG_FORMAT)
+            .Matches(@"^([a-zA-Z0-9-_]|%[A-F0-9]{2})*$")
+                .WithErrorCode(SLUG_FORMAT)
+                .WithMessage("Invalid slug format.");
```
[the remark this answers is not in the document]
TheTedder: I brought this up before and basically we have a few options.
|  | Separate Entities/Models | Combined Entities/Models |
| --- | :---: | --- |
| **Validate data as we get it** | | |
| **Validate data in the db**| | |
| **Validate on both ends** | (our setup) | |

Validating data before it even gets to the database is something that I want to keep doing since it improves performance. If you can terminate the request before it even makes a connection to the database then response times will improve. On the other hand, you definitely want to validate on the database side as well so that garbage data that maybe didn't come in through the backend doesn't get persisted. So the only thing that would help then is to merge our DB entities and our DTOs together and simply use annotations and/or fluent validation to control them. I am not opposed to this idea but I think it'll have to be a separate PR.

TheTedder:
I have removed support for URL-encoded characters in slugs because it didn't make much sense anyway.

LeaderboardBackend/Models/Validation/SlugRule.cs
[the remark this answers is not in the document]
zysim: I'll have to see examples of merges, cuz the way I imagine things, you will have fields in DTOs that won't exist in entities and vice versa, e.g. response DTOs will have fields taken from relations. But okay; thanks for the reminder on why we validate this way.

zysim (approved):

PR #232, opened 2024-08-18
Get Leaderboard Endpoint

### LINKED ISSUE #228: Get Leaderboard Endpoint
`GET` `/api/leaderboard/{id}`.

If `id` is not a long, the server must return a 404 error.

If the given ID does not correspond to a leaderboard in the database, the server must return a 404 error.

If the given ID corresponds to a leaderboard in the database, the server must return a 200 OK status and a JSON response body containing an object with the following fields:
- Id: long
- Name: string
- Slug: string
- Info: string?
- CreatedAt: timestamp
- UpdatedAt: timestamp?
- DeletedAt: timestamp?

Leaderboards that are deleted may be returned.

zysim:
Closes #228

### DIFF
--- a/LeaderboardBackend.Test/Leaderboards.cs
+++ b/LeaderboardBackend.Test/Leaderboards.cs
@@ -38,11 +38,12 @@ internal class Leaderboards
     public void OneTimeTearDown() => _factory.Dispose();
 
     [Test]
-    public void GetLeaderboard_NotFound() => _apiClient.Invoking(a => a.Get<LeaderboardViewModel>(
-        $"/api/leaderboard/{long.MaxValue}",
-        new()
-    )).Should()
-    .ThrowAsync<RequestFailureException>()
+    public async Task GetLeaderboard_NotFound() => await FluentActions.Awaiting(
+        async () => await _apiClient.Get<LeaderboardViewModel>(
+            $"/api/leaderboards/{long.MaxValue}",
+            new()
+        )
+    ).Should().ThrowAsync<RequestFailureException>()
     .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
 
     [Test]
--- a/LeaderboardBackend.Test/Leaderboards.cs
+++ b/LeaderboardBackend.Test/Leaderboards.cs
@@ -40,7 +40,7 @@ internal class Leaderboards
     [Test]
     public async Task GetLeaderboard_NotFound() => await FluentActions.Awaiting(
         async () => await _apiClient.Get<LeaderboardViewModel>(
-            $"/api/leaderboards/{long.MaxValue}",
+            $"/api/leaderboard/{long.MaxValue}",
             new()
         )
     ).Should().ThrowAsync<RequestFailureException>()
--- a/LeaderboardBackend.Test/Leaderboards.cs
+++ b/LeaderboardBackend.Test/Leaderboards.cs
@@ -35,24 +35,15 @@ internal class Leaderboards
     }
 
     [OneTimeTearDown]
-    public void OneTimeTearDown()
-    {
-        _factory.Dispose();
-    }
+    public void OneTimeTearDown() => _factory.Dispose();
 
     [Test]
-    public void GetLeaderboard_NotFound()
-    {
-        RequestFailureException e = Assert.ThrowsAsync<RequestFailureException>(
-            async () =>
-                await _apiClient.Get<LeaderboardViewModel>(
-                    $"/api/leaderboard/{long.MaxValue}",
-                    new()
-                )
-        )!;
-
-        Assert.AreEqual(HttpStatusCode.NotFound, e.Response.StatusCode);
-    }
+    public void GetLeaderboard_NotFound() => _apiClient.Invoking(a => a.Get<LeaderboardViewModel>(
+        $"/api/leaderboard/{long.MaxValue}",
+        new()
+    )).Should()
+    .ThrowAsync<RequestFailureException>()
+    .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
 
     [Test]
     public async Task CreateLeaderboard_GetLeaderboard_OK()
--- a/LeaderboardBackend/Controllers/LeaderboardsController.cs
+++ b/LeaderboardBackend/Controllers/LeaderboardsController.cs
@@ -13,10 +13,8 @@ public class LeaderboardsController : ApiController
 {
     private readonly ILeaderboardService _leaderboardService;
 
-    public LeaderboardsController(ILeaderboardService leaderboardService)
-    {
+    public LeaderboardsController(ILeaderboardService leaderboardService) =>
         _leaderboardService = leaderboardService;
-    }
 
     [AllowAnonymous]
     [HttpGet("api/leaderboard/{id:long}")]
--- a/LeaderboardBackend/Controllers/LeaderboardsController.cs
+++ b/LeaderboardBackend/Controllers/LeaderboardsController.cs
@@ -9,12 +9,9 @@ using Swashbuckle.AspNetCore.Annotations;
 
 namespace LeaderboardBackend.Controllers;
 
-public class LeaderboardsController : ApiController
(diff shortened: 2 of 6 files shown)


LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -35,24 +35,15 @@ public async Task OneTimeSetUp()
...
                     $"/api/leaderboards/{long.MaxValue}",
                     new()
-                )
-        )!;
-
-        Assert.AreEqual(HttpStatusCode.NotFound, e.Response.StatusCode);
-    }
+                )).Should()
+                .ThrowAsync<RequestFailureException>()
+                .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
```
TheTedder, line 46: This needs to be awaited.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -12,17 +12,14 @@ public LeaderboardService(ApplicationContext applicationContext)
         _applicationContext = applicationContext;
     }
 
-    public async Task<Leaderboard?> GetLeaderboard(long id)
-    {
-        return await _applicationContext.Leaderboards.FindAsync(id);
-    }
+    public async Task<Leaderboard?> GetLeaderboard(long id) =>
+        await _applicationContext.Leaderboards
+            .FirstOrDefaultAsync(board => board.Id == id);
```
TheTedder, line 17: What's wrong with `FindAsync`? It's more a more concise way to write the same thing, isn't it?

zysim: I just wasn't aware of it, yeah

LeaderboardBackend.Test/Leaderboards.cs
TheTedder: The chained call to `Where` returns a `Task` that needs to be awaited. See https://fluentassertions.com/exceptions/.

The correct code is something like
```cs
await FluentActions.Awaiting(async () => await _apiClient.Get<LeaderboardViewModel>(
    $"/api/leaderboards/{long.MaxValue}",
    new()
)).Should().ThrowAsync<RequestFailureException>()
.Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
```

Alternatively, use a `Func`:
```cs
Func<Task<LeaderboardViewModel>> act = async () => await _apiClient.Get<LeaderboardViewModel>(
    $"/api/leaderboards/{long.MaxValue}",
    new()
);

await act.Should().ThrowAsync<RequestFailureException>()
.Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
```

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -3,26 +3,18 @@
 
 namespace LeaderboardBackend.Services;
 
-public class LeaderboardService : ILeaderboardService
+public class LeaderboardService(ApplicationContext applicationContext) : ILeaderboardService
 {
-    private readonly ApplicationContext _applicationContext;
+    private readonly ApplicationContext _applicationContext = applicationContext;
```
TheTedder, line 8: Don't need to store the parameter when using a primary constructor. See https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#primary-constructors.

TheTedder (changes requested):

LeaderboardBackend.Test/Leaderboards.cs
zysim: Oh whoops. Didn't realise that my test was passing for me because the result wasn't even resolved.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -3,26 +3,18 @@
...
 
-    public LeaderboardService(ApplicationContext applicationContext)
-    {
-        _applicationContext = applicationContext;
-    }
+    public ValueTask<Leaderboard?> GetLeaderboard(long id) =>
+        _applicationContext.Leaderboards
+            .FindAsync([id]);
```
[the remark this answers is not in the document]
zysim: Yeah I read the same post, too, after you suggested `FindAsync`. We really didn't need to change methods after all

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -35,24 +35,15 @@ public async Task OneTimeSetUp()
...
-                    new()
-                )
-        )!;
-
-        Assert.AreEqual(HttpStatusCode.NotFound, e.Response.StatusCode);
-    }
+    public async Task GetLeaderboard_NotFound() => await _apiClient.Awaiting(a => a.Get<LeaderboardViewModel>(
+        $"/api/leaderboard/{long.MaxValue}",
+        new()
+    )).Should()
```
TheTedder, line 44: I would suggest using an `async` lambda and awaiting the call.

LeaderboardBackend/Controllers/LeaderboardsController.cs
```diff
@@ -9,14 +9,9 @@
...
-public class LeaderboardsController : ApiController
+public class LeaderboardsController(ILeaderboardService leaderboardService) : ApiController
 {
-    private readonly ILeaderboardService _leaderboardService;
-
-    public LeaderboardsController(ILeaderboardService leaderboardService)
-    {
-        _leaderboardService = leaderboardService;
-    }
+    private readonly ILeaderboardService _leaderboardService = leaderboardService;
```
TheTedder, line 14: You don't need to store the parameter in a field.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -3,45 +3,35 @@
...
-        return await _applicationContext.Leaderboards.FindAsync(id);
-    }
-
-    public async Task<Leaderboard?> GetLeaderboardBySlug(string slug)
-    {
-        return await _applicationContext.Leaderboards
+    public Task<Leaderboard?> GetLeaderboardBySlug(string slug) =>
+        applicationContext.Leaderboards
             .AsNoTracking()
             .FirstOrDefaultAsync(x => x.Slug == slug);
```
TheTedder, line 15: I would suggest awaiting this immediately.

TheTedder (commented):
Some minor stuff. I didn't have much time today so it's not a super thorough review. I'll potentially add more comments if I think of them later.

LeaderboardBackend/Controllers/LeaderboardsController.cs
zysim: Eh

Guess I forgot to remove this

LeaderboardBackend.Test/Leaderboards.cs
zysim: But the test works right now so why would we want to switch to a lambda? That being said; how would I do that? Like this?

```cs
public void GetLeaderboard_NotFound() {
    Func<Task> act = () => _apiClient.Awaiting(a => a.Get<LeaderboardViewModel>(
        $"/api/leaderboard/{long.MaxValue}",
        new()
    ))
    await act.Should().ThrowAsync<RequestFailureException>()
        .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
}
```

TheTedder: No I just mean await the call to Get using an async anonymous function. Just a suggestion.

zysim: Sure, but jsyk it's not actually needed; it behaves correctly. Even the example in the docs doesn't explicitly await it:
```cs
// From https://fluentassertions.com/exceptions/
Func<Task> act = asyncObject.Awaiting(x => x.ThrowAsync<ArgumentException>());
await act.Should().ThrowAsync<ArgumentException>();
```

TheTedder (approved):

PR #233, opened 2024-08-18
Get Leaderboard By Slug

### LINKED ISSUE #230: Get Leaderboard by Slug Endpoint
`GET` `/api/leaderboard?slug={slug}`.

If `slug` does not correspond to a leaderboard in the database, the server must return a 404 error.

If `slug` corresponds to a deleted leaderboard in the database, the server must return a 404 error.

If the given slug corresponds to a non-deleted leaderboard in the database, the server must return a 200 OK status and a JSON response body containing an object with the following fields:
- Id: long
- Name: string
- Slug: string
- Info: string?
- CreatedAt: timestamp
- UpdatedAt: timestamp?
- DeletedAt: timestamp?

zysim:
Closes #230


TheTedder (changes requested):
Make it so that deleted leaderboards are not returned. Add a test case or two for this scenario.

TheTedder (approved):

PR #239, opened 2024-08-21
Issue 235: Added [Required] to all occurrences of Name.

yash-saini:


TheTedder:
Actually, only the actual DB entities need to have it. We're using Fluent Validation for our request DTOs. Should have been a bit clearer in the issue description. Additionally, make a new migration so that these changes are reflected in the database schema.

TheTedder:
Addressed in #242

PR #242, opened 2024-08-31
DB Model Updates

### LINKED ISSUE #235: Task: Disallow Blank Names
Currently, name fields are marked non-nullable but there is nothing stopping them from being empty strings.

The fix: put a `[Required]` annotation on the field. The model builder will interpret this as disallowing "empty" values.

TheTedder:
- updated EF tool and packages
- switched from non-deterministic collation to `citext`
- added missing check constraints
- filtered deleted LBs from slug index

Closes #235

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "dotnet-ef": {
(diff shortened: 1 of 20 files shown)


Eein (approved):
Everything looks good from my end

zysim (approved):

PR #244, opened 2024-09-06
Automatically Set CreatedAt and UpdatedAt

### LINKED ISSUE #237: Automatically Set `UpdatedAt` Fields
Make it so that the `UpdatedAt` fields of various models automatically get set to the current time when any of the object's fields change. Read this article for some inspiration: https://learn.microsoft.com/en-us/ef/core/logging-events-diagnostics/events#example-timestamp-state-changes.

TheTedder:
What it says on the tin. Had to mess with the tests a bit to get this working properly. Also removes erroneously overridden `Equals` and `GetHashCode` on models (this is a no-no). Closes #237

### DIFF
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -6,6 +6,11 @@ using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Models.ViewModels;
 using LeaderboardBackend.Test.TestApi;
 using LeaderboardBackend.Test.TestApi.Extensions;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using NodaTime;
+using NodaTime.Testing;
 using NUnit.Framework;
 
 namespace LeaderboardBackend.Test;
@@ -14,16 +19,23 @@ namespace LeaderboardBackend.Test;
 internal class Categories
 {
     private static TestApiClient _apiClient = null!;
-    private static TestApiFactory _factory = null!;
+    private static WebApplicationFactory<Program> _factory = null!;
+    private static readonly FakeClock _clock = new(new Instant());
     private static string? _jwt;
 
     [OneTimeSetUp]
     public async Task OneTimeSetUp()
     {
-        _factory = new TestApiFactory();
-        _apiClient = _factory.CreateTestApiClient();
+        _factory = new TestApiFactory().WithWebHostBuilder(builder =>
+        {
+            builder.ConfigureTestServices(services =>
+            {
+                services.AddSingleton<IClock, FakeClock>(_ => _clock);
+            });
+        });
+        _apiClient = new TestApiClient(_factory.CreateClient());
 
-        _factory.ResetDatabase();
+        PostgresDatabaseFixture.ResetDatabaseToTemplate();
         _jwt = (await _apiClient.LoginAdminUser()).Token;
     }
 
@@ -44,6 +56,9 @@ internal class Categories
     [Test]
     public static async Task CreateCategory_GetCategory_OK()
     {
+        Instant now = Instant.FromUnixTimeSeconds(1);
+        _clock.Reset(now);
+
         LeaderboardViewModel createdLeaderboard = await _apiClient.Post<LeaderboardViewModel>(
             "/leaderboards/create",
             new()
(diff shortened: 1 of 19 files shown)


zysim:
Kinda don't like having inheritance on the interfaces themselves. If we do go down with the interfaces approach; I'd rather we do something like this:
```cs
public class Leaderboard : ICreatedAt, IUpdatedAt
```
Alternatively as I've said in the channel; we just do a base class with all three timestamps, and have classes inherit that

TheTedder:
> ~Kinda don't like having inheritance on the interfaces themselves. If we do go down with the interfaces approach; I'd rather we do something like this:~ (actually it really doesn't matter. Though I would still prefer if we did the base class, though)
> 
> ```cs
> public class Leaderboard : ICreatedAt, IUpdatedAt
> ```
> 
> Alternatively as I've said in the channel; we just do a base class with all three timestamps, and have classes inherit that

Having one base class with all three is impossible since some tables have CreatedAt but not UpdatedAt. I couldn't see an easy way to automate DeletedAt so I'm keeping that one manual for now. Would you still like to me get rid of the inheritance?

zysim:
Yeah nah just keep what you have. Don't have to that class inheritance thing.

LeaderboardBackend.Test/Features/Users/ConfirmAccountTests.cs
```diff
@@ -23,7 +23,13 @@ public class ConfirmAccountTests : IntegrationTestsBase
...
     public void Init()
     {
-        _scope = _factory.Services.CreateScope();
+        _scope = _factory.WithWebHostBuilder(builder =>
+        {
+            builder.ConfigureTestServices(services =>
+            {
+                services.AddSingleton<IClock, FakeClock>(_ => _clock);
+            });
+        }).Services.CreateScope();
```
zysim, line 32: Why can't we use the old line? Do we have to instantiate a new singleton every time? Plus why's the Categories test initialised with a new `TestApiFactory` while this one isn't?

zysim:
How'd you figure out on how to add methods in ApplicationContext like that?

LeaderboardBackend.Test/Categories.cs
```diff
@@ -58,27 +73,31 @@ public static async Task CreateCategory_GetCategory_OK()
...
-                    Info = null,
-                    SortDirection = SortDirection.Ascending,
-                    Type = RunType.Time
-                },
+                Body = request,
                 Jwt = _jwt
             }
         );
 
+        createdCategory.CreatedAt.Should().Be(now);
```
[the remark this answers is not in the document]
TheTedder: It actually only takes effect once you save it to the db.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -58,27 +73,31 @@ public static async Task CreateCategory_GetCategory_OK()
...
+
         CategoryViewModel retrievedCategory = await _apiClient.Get<CategoryViewModel>(
             $"/api/category/{createdCategory?.Id}", new() { }
         );
 
-        Assert.AreEqual(createdCategory, retrievedCategory);
+        retrievedCategory.Should().BeEquivalentTo(request, opts => opts.Excluding(c => c.LeaderboardId));
```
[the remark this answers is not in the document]
TheTedder: Have to exclude that field from the equivalence check because it only exists on one of the types of the objects being compared.

TheTedder:
> How'd you figure out on how to add methods in ApplicationContext like that?

I'm actually a wizard IRL. Also I read [this article](https://learn.microsoft.com/en-us/ef/core/logging-events-diagnostics/events#example-timestamp-state-changes).

zysim:
Would `IHasUpdateTimestamp` trigger on creation as well?

TheTedder:
> Would `IHasUpdateTimestamp` trigger on creation as well?

I'm not sure but it doesn't really matter either way as far as I'm concerned.

LeaderboardBackend.Test/Features/Users/ConfirmAccountTests.cs
TheTedder: > Why can't we use the old line?
Because we need to inject the fake clock.

> Do we have to instantiate a new singleton every time?
Nope, I was just moving quickly and this was the first way I thought to do it. I can change it if you want. That being said, it's reusing the same FakeClock every time so it's not instantiating a new one.

> Plus why's the Categories test initialised with a new TestApiFactory while this one isn't?
Because the Categories test class doesn't inherit from IntegrationTestsBase, which has its own factory member that I can't overwrite. I could take the take to normalize everything but I didn't because I just wanted to get this done. I could go back and change all the test classes to be set up the same way if you want me to.

zysim: Yeah let's just get this done. No need to change anything. I'll resolve this

zysim (approved):

PR #246, opened 2024-09-22
Add Test for Filtered Index

TheTedder:
This PR simply adds a test to make sure that deleted boards aren't hogging slugs.

### DIFF
--- a/LeaderboardBackend.Test/Leaderboards.cs
+++ b/LeaderboardBackend.Test/Leaderboards.cs
@@ -183,8 +183,8 @@ internal class Leaderboards
 
         Leaderboard deletedBoard = new()
         {
-            Name = "Super Mario 64 (OLD)",
-            Slug = "super-mario-64",
+            Name = "Super Mario World (OLD)",
+            Slug = "super-mario-world",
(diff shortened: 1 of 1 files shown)


LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -165,18 +165,53 @@ public async Task GetLeaderboards_Deleted_BySlug_NotFound()
...
+        Leaderboard deletedBoard = new()
+        {
+            Name = "Super Mario World (OLD)",
+            Slug = "super-mario-world",
+            DeletedAt = _clock.GetCurrentInstant()
+        };
+
+        context.Leaderboards.Add(deletedBoard);
+        await context.SaveChangesAsync();
+        deletedBoard.Id.Should().NotBe(default);
```
zysim, line 193: What does default resolve to? new Guid()?

zysim (approved):

LeaderboardBackend.Test/Leaderboards.cs
TheTedder: It's a `long` so zero. This is just a more idiosyncratic way of expressing it.

PR #248, opened 2024-09-29
Create LB Endpoint

### LINKED ISSUE #223: Create Leaderboard Endpoint
`POST` `/leaderboards/create`.

The request body includes the following fields:
- Name: string
- Slug: string
- Info: string?

The server returns a 401 error if the client is not authenticated.

The server returns a 403 error if the current user's role is not `ADMIN`.

The server returns a 409 error if the provided slug is already in use.

The server returns a 422 error if 
- slug is not between 2 and 80 characters inclusive.
- the slug is not comprised solely of alphanumeric digits, hyphens, and underscores.
- Name is absent or null.
- Slug is absent or null.

Upon succesful leaderboard creation, the server returns 201 CREATED status and a response body that includes the newly-created leaderboard as well as its location on the server.
```[tasklist]
### Tasks
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/214
- [ ] https://github.com/leaderboardsgg/leaderboard-backend/issues/226
```

TheTedder:
Updates LB creation endpoint to be in line with the spec. Closes #223.
Also changes up some code around request object validation to handle more cases and just do stuff a little more elegantly.

### DIFF
--- a/LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
+++ b/LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
@@ -1,7 +1,10 @@
+using System.ComponentModel.DataAnnotations;
 using FluentValidation;
 using LeaderboardBackend.Models.Validation;
 
 namespace LeaderboardBackend.Models.Requests;
+
+#nullable disable warnings
 
 /// <summary>
 ///     This request object is sent when creating a `Leaderboard`.
@@ -12,18 +15,21 @@ public record CreateLeaderboardRequest
     ///     The display name of the `Leaderboard` to create.
     /// </summary>
     /// <example>Foo Bar</example>
-    public required string Name { get; set; }
+    [Required]
+    public string Name { get; set; }
 
     /// <summary>
     ///     The URL-scoped unique identifier of the `Leaderboard`.<br/>
     ///     Must be [2, 80] in length and consist only of alphanumeric characters and hyphens.
     /// </summary>
     /// <example>foo-bar</example>
-    public required string Slug { get; set; }
+    [Required]
+    public string Slug { get; set; }
 
     /// <inheritdoc cref="Entities.Leaderboard.Info" />
-    public required string? Info { get; set; }
+    public string? Info { get; set; }
 }
+
 public class CreateLeaderboardRequestValidator : AbstractValidator<CreateLeaderboardRequest>
 {
     public CreateLeaderboardRequestValidator()
--- a/LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
+++ b/LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
@@ -26,5 +26,9 @@ public record CreateLeaderboardRequest
 }
 public class CreateLeaderboardRequestValidator : AbstractValidator<CreateLeaderboardRequest>
 {
-    public CreateLeaderboardRequestValidator() => RuleFor(x => x.Slug).Slug();
+    public CreateLeaderboardRequestValidator()
+    {
+        RuleFor(x => x.Name).NotEmpty();
+        RuleFor(x => x.Slug).Slug();
+    }
 }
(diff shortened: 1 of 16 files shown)


LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
```diff
@@ -1,8 +1,11 @@
+using System.ComponentModel.DataAnnotations;
 using FluentValidation;
 using LeaderboardBackend.Models.Validation;
 
 namespace LeaderboardBackend.Models.Requests;
 
+#nullable disable warnings
```
zysim, line 7: Okay so if I understand this line correctly; we're disabling the compiler warnings that the DTO fields may be nullable, because we _can_ expect them to be, right?

LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
```diff
@@ -12,19 +15,26 @@ public record CreateLeaderboardRequest
     ///     The display name of the `Leaderboard` to create.
     /// </summary>
     /// <example>Foo Bar</example>
-    public required string Name { get; set; }
+    [Required]
```
zysim, line 18: Why are we using the attribute here? Does it actually validate after all?

LeaderboardBackend/Program.cs
```diff
@@ -136,7 +136,9 @@
         // Enforces JSON output and causes OpenAPI UI to correctly show that we return JSON.
         opt.OutputFormatters.RemoveType<StringOutputFormatter>();
         opt.ModelBinderProviders.Insert(0, new UrlSafeBase64GuidBinderProvider());
+        opt.Filters.AddService<ValidationFilter>();
```
zysim, line 139: This is for my own learning; do we need to call `AddSingleton` (or basically add the service to `builder.Services` as you've done below whenever we add a new entry here?

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -27,9 +30,26 @@ public async Task<List<Leaderboard>> GetLeaderboards(long[]? ids = null)
...
+            Info = request.Info
+        };
+
+        applicationContext.Leaderboards.Add(lb);
+
+        try
+        {
+            await applicationContext.SaveChangesAsync();
+        }
+        catch (DbUpdateException e) when (e.InnerException is PostgresException { SqlState: PostgresErrorCodes.UniqueViolation })
```
zysim, line 48: What happens with any other exceptions thrown from the DB?

LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
TheTedder: For the swagger doc

LeaderboardBackend/Services/Impl/LeaderboardService.cs
TheTedder: 500 error. In theory that really shouldn't happen.

LeaderboardBackend/Program.cs
TheTedder: Only if it's a service

LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
TheTedder: The fact that we have to put this here anyway for swagger is one of the reasons I want to just use data annotations instead of FluentValidation.

LeaderboardBackend/Models/Requests/LeaderboardRequests.cs
TheTedder: Yes. We want to allow the props to implicitly default to null but to then fail validation. Otherwise it fails to bind at all with missing props and returns 400. It's a bit gross but so is the way c# does nullability in general.

zysim (approved):

PR #249, opened 2024-10-01
List Leaderboards Endpoint

### LINKED ISSUE #227: List Leaderboards Endpoint
`GET` `/api/leaderboards`.

The server must return a 200 OK status.

The server must return a JSON response body containing an array of leaderboard objects that have the following fields:
- Id: long
- Name: string
- Slug: string
- Info: string?
- CreatedAt: timestamp
- UpdatedAt: timestamp?
- DeletedAt: timestamp?

Leaderboards that are deleted must not be returned.

TheTedder:
This PR essentially splits the get leaderboards endpoint into two. If you provide IDs, it'll return only the Leaderboards with those IDs; otherwise, it'll return all non-deleted Leaderboards. Closes #227

### DIFF
--- a/LeaderboardBackend/Models/ViewModels/LeaderboardViewModel.cs
+++ b/LeaderboardBackend/Models/ViewModels/LeaderboardViewModel.cs
@@ -54,14 +54,14 @@ public record LeaderboardViewModel
     /// <summary>
     ///     A collection of `Category` entities for the `Leaderboard`.
     /// </summary>
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public IList<CategoryViewModel> Categories { get; init; }
+    public required IList<CategoryViewModel> Categories { get; init; }
 
     public static LeaderboardViewModel MapFrom(Leaderboard leaderboard)
     {
         IList<CategoryViewModel> categories = leaderboard.Categories
             ?.Select(CategoryViewModel.MapFrom)
             .ToList();
+
         return new LeaderboardViewModel
         {
             Id = leaderboard.Id,
@@ -71,7 +71,7 @@ public record LeaderboardViewModel
             CreatedAt = leaderboard.CreatedAt,
             UpdatedAt = leaderboard.UpdatedAt,
             DeletedAt = leaderboard.DeletedAt,
-            Categories = categories,
+            Categories = categories ?? Array.Empty<CategoryViewModel>(),
         };
     }
 }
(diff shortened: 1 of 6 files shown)


LeaderboardBackend/Models/ViewModels/LeaderboardViewModel.cs
```diff
@@ -51,11 +54,12 @@ public record LeaderboardViewModel
     /// <summary>
     ///     A collection of `Category` entities for the `Leaderboard`.
     /// </summary>
-    public required IList<CategoryViewModel> Categories { get; init; }
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
```
zysim, line 57: With this change though; we'll have to ensure we do the same thing for any future endpoint with possibly-null relations, for consistency. Imo it's fine to default to an empty array, anyway

TheTedder: All this does is not serialize it when there are no cats to be returned.

zysim: Ergo the field itself won't appear in the return object, right?

Eein (commented):
Seems good to me as long as theres a tracking issue for paginating the full list.

Also I noticed that categories are on the view model, if these are returned during the list view can you test the expected behaviour of these returning and being absent?

If they can never be returned in the leaderboards list, can you also test you should not find categories in this model?

LeaderboardBackend/Models/ViewModels/LeaderboardViewModel.cs
Eein: That could be a problem if frontend is expecting it.

I'd ask the frontend people what they would expect here.
Anecdotally the frontend teams I generally work with would expect an empty array. It just helps avoid safe access checks cluttering up the code, and having to write a new type for leaderboards with and without the field.

TheTedder: The frontend code that actually deals with the models is autogenerated so it'll mark the field as potentially absent. If they try to access it without doing a null check, it'll produce a warning.

TheTedder:
> Seems good to me as long as theres a tracking issue for paginating the full list.
> 
> Also I noticed that categories are on the view model, if these are returned during the list view can you test the expected behaviour of these returning and being absent? (this may be better in ViewModel tests if those exist; helps boilerplate down the road if there are special exceptions)

Categories are not implemented yet. We're gonna start worrying about them next "sprint" I believe so I'm just ignoring everything to do with them for now.

TheTedder:
I have decided it's best to remove the functionality of fetching multiple LBs at once by ID as it's best not to have endpoints we don't plan on using. That way we don't have to test and maintain them.

Eein (approved):
Ship it!

zysim (approved):

Issue #240, opened 2024-08-21
Restore Leaderboard Endpoint


TheTedder:
`PUT` `/leaderboard/{id}/restore` where {id} is a long.

The server must return a 401 error if the client is not authenticated.

The server must return a 403 error if the current user's role is not `ADMIN`.

The server must return a 404 error if the given ID does not correspond to a leaderboard in the database.

The server must return a 404 error if the given ID corresponds to a leaderboard that is not deleted.

The server may use a response body to disambiguate between the two different scenarios that can generate a 404 error.

The server must return a 409 error if the leaderboard cannot be restored because its slug is the same as an existing non-deleted leaderboard.

Upon a successful restoration, the server must set the `DeletedAt` field of the corresponding leaderboard to `null` and return a 204 status code.

zysim:
Actually; shouldn't a non-deleted board cause a 204 instead? Basically communicate a "we didn't have to do anything" instead of a "we couldn't find it".

zysim:
Will implement stuff as mentioned in the issue though, in the meantime.

Eein:
If the endpoint only takes deleted resource ids i think 404 is acceptable here

204 can be misconstrued that something occured

PR #250, opened 2024-10-14
Restore leaderboards

zysim:
Closes: #240 

~~I do _not_ know how to write these dang tests.~~ I now _do_ know how to write these dang tests.


Eein:
If leaderboard change is successful, frontend will probably want the leaderboard returned with the changes instead of a NoContent;

They will likely have to update a local cache/store - so this is better than making another request.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -339,6 +341,71 @@ public async Task DeletedBoardsDontConsumeSlugs()
...
+        HttpResponseMessage res = await _apiClient.Put($"/leaderboard/{deletedBoard.Id}/restore", new()
+        {
+            Jwt = _jwt
+        });
+
+        res.StatusCode.Should().Be(HttpStatusCode.NoContent);
+
+        Leaderboard? board = await context.Leaderboards.FindAsync(deletedBoard.Id);
+        board.Should().NotBeNull();
+        // TODO: `DeletedAt` is still not null here. Don't know how to fix it.
```
Eein, line 369: I'm learning a bit too, and apparently when you want to test an assertion, you must pull an entirely new database context - ie:
```cs
        ApplicationContext verifyContext = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();
        Leaderboard? board = await verifyContext.Leaderboards.FindAsync(deletedBoard.Id);
```

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -339,6 +341,71 @@ public async Task DeletedBoardsDontConsumeSlugs()
...
 
+    [Test]
+    public async Task RestoreLeaderboard_OK()
+    {
+        ApplicationContext context = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();
+
+        Leaderboard deletedBoard = new()
+        {
+            Name = "Super Mario World",
+            Slug = "super-mario-world-deleted",
```
Eein, line 352: you'll need to set a unique slug here... 
The tests aren't transactional, so data is persisting - i think theres already been some discussion on how to fix this. I've looked, but its quite complicated in ASP.net to set this up...

zysim: Yeah I realised they weren't at the time already; just forgot about it pretty much. I've simply swapped slugs since (wow the alliteration). And yeah IMO it's plain not worth the hassle to set that up for now.

LeaderboardBackend.Test/Leaderboards.cs
zysim: Thankfully we can just clear the change tracker as Ted's said in the chat. Much smaller word salad too. I've updated.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -14,6 +15,7 @@
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.VisualBasic;
```
TheTedder, line 18: Accidental import.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +350,107 @@ public async Task GetLeaderboards()
...
+
+    [Test]
+    public async Task RestoreLeaderboard_NotFound_WasNeverDeleted()
+    {
+        ApplicationContext context = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();
+
+        Leaderboard board = new()
+        {
+            Name = "Super Mario World",
+            Slug = "super-mario-world-non-deleted",
```
TheTedder, line 440: Super Mario World is used elsewhere in the tests. For clarity, pick a different one.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -42,4 +43,28 @@ public async Task<CreateLeaderboardResult> CreateLeaderboard(CreateLeaderboardRe
...
+        }
+
+        if (lb.DeletedAt == null)
+        {
+            return new LeaderboardNeverDeleted();
+        }
+
+        applicationContext.Leaderboards.Update(lb);
+
+        lb.UpdatedAt = clock.GetCurrentInstant();
```
TheTedder, line 63: If this is broken we should fix it instead of just doing it manually. I can look at it later.

LeaderboardBackend/Controllers/LeaderboardsController.cs
```diff
@@ -86,4 +86,28 @@ public async Task<ActionResult<LeaderboardViewModel>> CreateLeaderboard(
...
+    {
+        RestoreLeaderboardResult r = await leaderboardService.RestoreLeaderboard(id);
+
+        return r.Match<ActionResult<LeaderboardViewModel>>(
+            board => Ok(LeaderboardViewModel.MapFrom(board)),
+            notFound => NotFound(),
+            neverDeleted =>
+            {
+                ModelState.AddModelError("Leaderboard", "LeaderboardWasNeverPreviouslyDeleted");
+                return NotFound(new ValidationProblemDetails(ModelState));
```
TheTedder, line 109: I'm not sure that this counts as a validation error. Maybe just make a new DTO for this?

TheTedder (changes requested):
I think we forgot to cover the case where the board can't be restored because its slug has been reclaimed by another board. Shall we simply have it return a 409 and possibly the conflicting board or at least the ID of the conflicting board?

zysim:
> I think we forgot to cover the case where the board can't be restored because its slug has been reclaimed by another board. Shall we simply have it return a 409 and possibly the conflicting board or at least the ID of the conflicting board?

409'ing sounds fair. I should add that in.

LeaderboardBackend.Test/Leaderboards.cs
zysim: Dangit.

Eein:
> > I think we forgot to cover the case where the board can't be restored because its slug has been reclaimed by another board. Shall we simply have it return a 409 and possibly the conflicting board or at least the ID of the conflicting board?
> 
> 409'ing sounds fair. I should add that in. I'll return the full board; better visibility that way.

Is this even possible?
If a leaderboard is deleted, it should still retain the unique index no?

TheTedder:
> > > I think we forgot to cover the case where the board can't be restored because its slug has been reclaimed by another board. Shall we simply have it return a 409 and possibly the conflicting board or at least the ID of the conflicting board?
> > 
> > 409'ing sounds fair. I should add that in. I'll return the full board; better visibility that way.
> 
> Is this even possible?
> If a leaderboard is deleted, it should still retain the unique index no?
> 

We decided that deleted LBs should get excluded from the unique index for slugs. This is achieved by filtering the index to only cover LBs that are not deleted. This means that when you delete an LB, its slug is freed. Alternatively, we could ask for a slug to use when restoring the LB to prevent the annoying and confusing case of having to give a deleted LB new slug _before_ you restore it. In addition or perhaps instead, we could have it so that deleting a board automatically clears its slug.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +350,107 @@ public async Task GetLeaderboards()
...
+            Jwt = jwt,
+        });
+
+        await act.Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.Forbidden);
+    }
+
+    [Test]
+    public async Task RestoreLeaderboard_NotFound()
+    {
+        Func<Task<LeaderboardViewModel>> act = async () => await _apiClient.Put<LeaderboardViewModel>($"/leaderboard/100/restore", new()
```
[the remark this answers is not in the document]
zysim: 1e10 it is

LeaderboardBackend/Controllers/LeaderboardsController.cs
zysim: Frankly I only did it cuz I was copying existing tests. Currently made it so that it just returns a string saying "Was never deleted". Feel free to suggest something better? I can't imagine what else to say here.

zysim:
Reflecting what I've said on Discord; I won't be implementing the ability to pass a new slug in to restore a board, nor am I clearing a board's slug on deletion.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +346,104 @@ public async Task GetLeaderboards()
...
+        _clock.AdvanceMinutes(1);
+
+        LeaderboardViewModel res = await _apiClient.Put<LeaderboardViewModel>($"/leaderboard/{deletedBoard.Id}/restore", new()
+        {
+            Jwt = _jwt
+        });
+
+        res.Id.Should().Be(deletedBoard.Id);
+        res.Slug.Should().Be(deletedBoard.Slug);
+        res.UpdatedAt.Should().Be(res.CreatedAt + Duration.FromMinutes(1));
```
TheTedder, line 375: Might be better to just check that UpdatedAt is equal to the current time. That way, a bug with CreatedAt won't end up affecting this test.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +346,104 @@ public async Task GetLeaderboards()
...
+        res.DeletedAt.Should().BeNull();
+    }
+
+    [Test]
+    public async Task RestoreLeaderboard_Unauthenticated()
+    {
+        Func<Task<LeaderboardViewModel>> act = async () => await _apiClient.Put<LeaderboardViewModel>($"/leaderboard/100/restore", new()
+        {
+            Jwt = ""
+        });
```
TheTedder, line 385: You can actually just not specify this value at all.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +346,104 @@ public async Task GetLeaderboards()
...
+
+        await act.Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.Unauthorized);
+    }
+
+    [TestCase("restore-leaderboard-unauth1@example.com", "RestoreBoard1", UserRole.Confirmed)]
+    [TestCase("restore-leaderboard-unauth2@example.com", "RestoreBoard2", UserRole.Registered)]
```
TheTedder, line 391: Don't forget the BANNED role. We should test that too.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +346,104 @@ public async Task GetLeaderboards()
...
+    public async Task RestoreLeaderboard_Unauthorized(string email, string username, UserRole role)
+    {
+        UserViewModel userModel = await _apiClient.RegisterUser(username, email, "P4ssword");
+
+        ApplicationContext context = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();
+
+        User? user = await context.Users.FindAsync([userModel.Id]);
```
TheTedder, line 398: Don't need to use an array to pass args into this function because it's variadic.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +346,104 @@ public async Task GetLeaderboards()
...
+
+        context.Users.Update(user!);
+        user!.Role = role;
```
TheTedder, line 401: Does the call to Update() do anything? When you fetch using FindAsync, it should start tracking changes to the returned object automatically. It might be simpler overall to just make this all one step by creating the user in the database directly instead of calling RegisterUser, fetching, updating, and then saving.

LeaderboardBackend.Test/Leaderboards.cs
TheTedder: You may need to modify the user's role _after_ they log in for this to work.

LeaderboardBackend/Controllers/LeaderboardsController.cs
TheTedder: Check [this page](https://learn.microsoft.com/en-us/aspnet/core/web-api/handle-errors) for ways to handle client errors. I recommend returning a ProblemDetails with a custom Title of NotDeleted or something similar.

LeaderboardBackend/Models/Entities/ApplicationContext.cs
```diff
@@ -41,7 +41,7 @@ public ApplicationContext(DbContextOptions<ApplicationContext> options, IClock c
     {
         _clock = clock;
         ChangeTracker.Tracked += AddCreationTimestamp;
-        ChangeTracker.Tracked += SetUpdateTimestamp;
+        ChangeTracker.StateChanged += SetUpdateTimestamp;
```
TheTedder, line 44: This change should already be on main. Do you need to do another rebase?

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -42,4 +43,34 @@ public async Task<CreateLeaderboardResult> CreateLeaderboard(CreateLeaderboardRe
...
+        {
+            return new LeaderboardNeverDeleted();
+        }
+
+        Leaderboard? maybe = await applicationContext.Leaderboards.SingleOrDefaultAsync(board => board.Slug == lb.Slug && board.DeletedAt == null);
+
+        if (maybe != null)
+        {
+            return new RestoreLeaderboardConflict(maybe);
+        }
```
TheTedder, line 66: Another way to do this would be to simply try to restore it and return the offender if it fails due to a conflict.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -42,4 +43,34 @@ public async Task<CreateLeaderboardResult> CreateLeaderboard(CreateLeaderboardRe
...
+
+        applicationContext.Leaderboards.Update(lb);
```
TheTedder, line 68: I don't think this is necessary. You already fetched it so it should track automatically.

LeaderboardBackend/Controllers/LeaderboardsController.cs
```diff
@@ -86,4 +86,30 @@ public async Task<ActionResult<LeaderboardViewModel>> CreateLeaderboard(
...
+
+        return r.Match<ActionResult<LeaderboardViewModel>>(
+            board => Ok(LeaderboardViewModel.MapFrom(board)),
+            notFound => NotFound(),
+            neverDeleted =>
+            {
+                ModelState.AddModelError("Leaderboard", "LeaderboardWasNeverPreviouslyDeleted");
+                return NotFound("Was never deleted");
+            },
+            conflict => Conflict(conflict.Board)
```
TheTedder, line 112: This needs to be mapped to a viewmodel.

TheTedder (changes requested):

LeaderboardBackend.Test/Leaderboards.cs
zysim: Yeah I originally had that role in, but it threw at login. And ye I'd have to modify the role. Adding that in.

LeaderboardBackend.Test/Leaderboards.cs
zysim: Tried doing that but I keep getting "Invalid salt version". This is what I tried to do:
```cs
public async Task RestoreLeaderboard_Unauthorized(string email, string username, UserRole role)
{
    ApplicationContext context = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();

    context.Users.Add(new()
    {
        Email = email,
        Password = "P4ssword",
        Username = username,
        Role = role
    });

    await context.SaveChangesAsync();
```

LeaderboardBackend/Models/Entities/ApplicationContext.cs
zysim: Guess so? I already did a rebase; guess it didn't go through or something.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
zysim: Can we just keep this? Unless it really becomes a problem.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
zysim: Oh yeah you're right

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -430,20 +469,70 @@ public async Task RestoreLeaderboard_NotFound_WasNeverDeleted()
...
+                {
+                    Jwt = _jwt,
+                }
+            );
+        }
+        catch (RequestFailureException e)
+        {
+            e.Response.StatusCode.Should().Be(HttpStatusCode.NotFound);
+            string? model = await e.Response.Content.ReadAsStringAsync();
+            model!.Should().MatchRegex("Not Deleted");
```
zysim, line 494: Re: this bit: tried to call `ReadFromJsonAsync` instead, but the resolved object didn't have the message at all; it was just `null`. I can't figure out how to get the response to serialise right.

LeaderboardBackend/Models/Entities/ApplicationContext.cs
zysim: Rebased and it's gone.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +349,190 @@ public async Task GetLeaderboards()
...
+
+        ApplicationContext context = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();
+
+        context.Users.Update(new()
+        {
+            Id = userModel.Id,
+            Role = UserRole.Banned,
+            Username = userModel.Username,
+            Email = email,
+            Password = password
```
TheTedder, line 412: You're actually unintentionally overwriting the user's salted and hashed password with a plaintext password here. It's not causing an issue right now because of the password you've chosen but it could in the future. Do something like `context.Users.Single(user => user.Id == userModel.Id).Role = UserRole.Banned;` instead.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +349,190 @@ public async Task GetLeaderboards()
...
+                }
+            );
+        }
+        catch (RequestFailureException e)
+        {
+            e.Response.StatusCode.Should().Be(HttpStatusCode.Conflict);
+            LeaderboardViewModel? model = await e.Response.Content.ReadFromJsonAsync<LeaderboardViewModel>(TestInitCommonFields.JsonSerializerOptions);
+            model.Should().NotBeNull();
+            model!.Slug.Should().Be("conflicted-mario-world");
+        }
```
TheTedder, line 536: This test is written incorrectly because in the case that request succeeds, the test doesn't fail (bad). Check the discord for a better way to write this.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -42,4 +43,32 @@ public async Task<CreateLeaderboardResult> CreateLeaderboard(CreateLeaderboardRe
 
         return lb;
     }
+
+    public async Task<RestoreLeaderboardResult> RestoreLeaderboard(long id)
+    {
+        Leaderboard? lb = await applicationContext.Leaderboards.FindAsync([id]);
```
TheTedder, line 49: Don't need to put method args in an array here.

TheTedder (changes requested):

LeaderboardBackend.Test/Leaderboards.cs
zysim: Yeah I know that. Just wanted to show that it wasn't possible for some reason (I know the user account tests work). And yeah; the FindAsync does track. Will update

LeaderboardBackend/Services/Impl/LeaderboardService.cs
zysim: Oh yeah that's true. I'll update that later. Was really tired when I replied; didn't wanna spend even more time changing code around ha

LeaderboardBackend.Test/Leaderboards.cs
zysim: Remind me to not write stuff when I'm tired

LeaderboardBackend/Services/Impl/LeaderboardService.cs
zysim: @TheTedder Trying to follow the example of that UserService method rn; you know I could get the conflicting board from the PostgresException? Or would I still need to make a DB call?

zysim: Also; following that pattern; I'd make a `LeaderboardEntityTypeConfig.SLUG_UNIQUE_INDEX` in the entity file. Would I also need to change the `HasFilter` call in the builder definition, too? I.e. edit `LeaderboardEntityTypeConfig`'s definition to:
```cs
public class LeaderboardEntityTypeConfig : IEntityTypeConfiguration<Leaderboard>
{
    public const string SLUG_UNIQUE_INDEX = "ix_leaderboards_slug";

    public void Configure(EntityTypeBuilder<Leaderboard> builder) =>
        builder.HasIndex(l => l.Slug)
            .IsUnique()
            .HasDatabaseName(SLUG_UNIQUE_INDEX); // replaces current `HasFilter` call
}
```

TheTedder: I'll take a closer look later when I get home but you don't need to check for which index specifically caused the issue because there's only one unique index on the table, meaning the table doesn't need to be modified.

TheTedder: You could do it like this, avoiding an extra DB call if no conflict happens.
```cs
public async Task<RestoreLeaderboardResult> RestoreLeaderboard(long id)
{
    Leaderboard? lb = await applicationContext.Leaderboards.FindAsync(id);

    if (lb == null)
    {
        return new LeaderboardNotFound();
    }

    if (lb.DeletedAt == null)
    {
        return new LeaderboardNeverDeleted();
    }

    lb.DeletedAt = null;

    try
    {
        await applicationContext.SaveChangesAsync();
    }
    catch (DbUpdateException e)
        when(e.InnerException is PostgresException { SqlState: PostgresErrorCodes.UniqueViolation } pgEx)
    {
        Leaderboard conflict = await applicationContext.Leaderboards.SingleAsync(c => c.Slug == lb.Slug && c.DeletedAt == null);
        return new RestoreLeaderboardConflict(conflict);
    }

    return lb;
}
```

zysim: I think the `return` should be inside the `try`, right? Or more importantly, what's the behaviour if a different exception is thrown?

TheTedder: Whether the return statement is inside the try block or not doesn't really make a difference. If an unrelated exception is thrown, the error goes uncaught and the server returns a 500 status code.

zysim: It be updated

TheTedder (approved):

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -348,4 +350,176 @@ public async Task GetLeaderboards()
...
+                new()
+                {
+                    Jwt = _jwt,
+                }
+            )
+        ).Should().ThrowAsync<RequestFailureException>().Where(ex => ex.Response.StatusCode == HttpStatusCode.Conflict);
+
+        LeaderboardViewModel? model = await exAssert.Which.Response.Content.ReadFromJsonAsync<LeaderboardViewModel>(TestInitCommonFields.JsonSerializerOptions);
+        model.Should().NotBeNull();
+        model!.Slug.Should().Be("conflicted-mario-world");
```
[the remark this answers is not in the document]
zysim: I should actually check for the ID, instead.

PR #251, opened 2024-10-16
Update Commit Workflow

Eein:
- removes husky and refs to it
- adds workflow dispatch
- adds ci runs on pr to main
- splits linting and builds

Users may have to run `git config --unset core.hooksPath` to get rid of the hooks from their local.

Can also just go into `.\.git\hooks` and remove everything that isn't a `*.sample`


TheTedder (approved):

zysim:
@Eein Why are we removing Husky btw?

PR #253, opened 2024-10-19
Delete Leaderboard Endpoint

### LINKED ISSUE #231: Delete Leaderboard Endpoint
`DELETE` `/leaderboard/{id}` where `id` is a long.

The server must return a 401 error if the client is not authenticated.

The server must return a 403 error if the current user's role is not `ADMIN`.

The server must return a 404 error if the given ID does not correspond to a leaderboard in the database.

The server must return a 404 error if the given ID corresponds to a leaderboard that is already deleted.

The server may use a response body to disambiguate between the two different scenarios that can generate a 404 error.

Upon a successful deletion, the server must set the `DeletedAt` field of the corresponding leaderboard to the current time and return a 204 status code.

TheTedder:
Closes #231. I chose to make the return type of the method on the service not specific to LBs so that we can reuse it.

### DIFF
--- a/LeaderboardBackend.Test/Leaderboards.cs
+++ b/LeaderboardBackend.Test/Leaderboards.cs
@@ -400,14 +400,11 @@ internal class Leaderboards
         await userService.CreateUser(registerRequest);
         context.Leaderboards.Add(lb);
         await context.SaveChangesAsync();
-
-#pragma warning disable IDE0008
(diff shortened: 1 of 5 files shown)


Eein (approved):
Looks good!

If you feel like it maybe a test for invalid type (guid as id)

zysim (approved):

PR #266, opened 2025-01-06
Create Categories

### LINKED ISSUE #259: Create Category Endpoint
`POST` `/leaderboard/{id}/categories/create`.

The request body must include the following fields:
* Name: string
* Slug: string
* Info: string?
* SortDirection: 'Ascending' | 'Descending'
* RunType: 'Time' | 'Score'

If the client is not authenticated, the server must return a 401 error .

If the current user's role is not `ADMIN`, the server must return a 403 error.

If the specified ID does not correspond to a leaderboard in the database, the server must return a 404 error.

If the leaderboard identified by the specified ID already contains a non-deleted category with the specified slug, the server must return a 409 error and return the conflicting category in the response body.

If any of the following conditions are met, the server must return a 422 error.

- Slug is not between 2 and 80 characters inclusive.
- Slug is not comprised solely of alphanumeric digits, hyphens, and underscores.
- Name is absent or null.
- Slug is absent or null.
- SortDirection is absent or an invalid value.
- RunType is absent or an invalid value.

Upon successful category creation, the server must return a 201 CREATED status with a response body that includes the newly-created category as well as its location on the server.

zysim:
Closes #259 

Todos:
- [x] Rebase PR once #269's done, and update tests
- [ ] Invalid enum fields in requests 400 instead of 422 (e.g. `"sortDirection": "Ascendingg"` in the request body) (to do post-MVP)
- [x] ~~Enhance 422 test case to check against invalid enum values~~ (Out of scope of this PR; needs validation logic change, which needs research)

### DIFF
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -92,7 +92,7 @@ internal class Categories
             }
         );
 
-        createdCategory.CreatedAt.Should().BeEquivalentTo(_clock.GetCurrentInstant());
+        createdCategory.CreatedAt.Should().Be(_clock.GetCurrentInstant());
 
         CategoryViewModel retrievedCategory = await _apiClient.Get<CategoryViewModel>(
             $"/api/category/{createdCategory?.Id}", new() { }
@@ -261,18 +261,17 @@ internal class Categories
         problemDetails!.Conflicting.Should().BeEquivalentTo(created);
     }
 
-    [TestCase(null, "bad-data", null, null, HttpStatusCode.UnprocessableContent)]
-    [TestCase("Bad Data", null, null, null, HttpStatusCode.UnprocessableContent)]
-    [TestCase("Bad Request Invalid SortDirection", "invalid-sort-direction", "Invalid SortDirection", null, HttpStatusCode.BadRequest)]
-    [TestCase("Bad Request Invalid Type", "invalid-type", null, "Invalid Type", HttpStatusCode.BadRequest)]
-    public static async Task CreateCategory_BadData(string? name, string? slug, string? invalidSortDirection, string? invalidRunType, HttpStatusCode expectedCode)
+    [TestCase(null, "bad-data", SortDirection.Ascending, RunType.Score, HttpStatusCode.UnprocessableContent)]
+    [TestCase("Bad Data", null, SortDirection.Ascending, RunType.Score, HttpStatusCode.UnprocessableContent)]
+    [TestCase("Bad Request Invalid SortDirection", "invalid-sort-direction", "Invalid SortDirection", RunType.Score, HttpStatusCode.BadRequest)]
+    [TestCase("Bad Request Invalid Type", "invalid-type", SortDirection.Ascending, "Invalid Type", HttpStatusCode.BadRequest)]
+    public static async Task CreateCategory_BadData(string? name, string? slug, dynamic sortDirection, dynamic runType, HttpStatusCode expectedCode)
     {
         var request = new
         {
             Name = name,
-            Info = "",
-            SortDirection = invalidSortDirection ?? SortDirection.Ascending.ToString(),
-            Type = invalidRunType ?? RunType.Score.ToString(),
+            SortDirection = sortDirection,
+            Type = runType,
             Slug = slug,
         };
 
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -92,7 +92,7 @@ internal class Categories
             }
         );
 
-        createdCategory.CreatedAt.Should().Be(_clock.GetCurrentInstant());
+        createdCategory.CreatedAt.Should().BeEquivalentTo(_clock.GetCurrentInstant());
 
         CategoryViewModel retrievedCategory = await _apiClient.Get<CategoryViewModel>(
             $"/api/category/{createdCategory?.Id}", new() { }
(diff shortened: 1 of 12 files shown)


LeaderboardBackend/Services/Impl/CategoryService.cs
```diff
@@ -1,29 +1,48 @@
...
-        await _applicationContext.SaveChangesAsync();
-    }
+        try
+        {
+            await applicationContext.SaveChangesAsync();
+        }
+        catch (DbUpdateException e) when (e.InnerException is PostgresException { SqlState: PostgresErrorCodes.ForeignKeyViolation })
+        {
+            return new NotFound();
+        }
```
zysim, line 37: Shit don't work. It only triggers 500s here, and I don't know how to fix this..

LeaderboardBackend/Services/Impl/CategoryService.cs
```diff
@@ -1,29 +1,48 @@
...
+        catch (DbUpdateException e) when (e.InnerException is PostgresException { SqlState: PostgresErrorCodes.UniqueViolation } pgEx)
+        {
+            return new Conflict<Category>();
```
TheTedder, line 40: Make it actually fetch the conflicting category.

TheTedder (changes requested):

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -32,29 +33,29 @@ public async Task<ActionResult<CategoryViewModel>> GetCategory(long id)
     [HttpPost("categories/create")]
     [SwaggerOperation("Creates a new Category. This request is restricted to Moderators.", OperationId = "createCategory")]
     [SwaggerResponse(201)]
-    [SwaggerResponse(403)]
-    [SwaggerResponse(422, Type = typeof(ValidationProblemDetails))]
+    [SwaggerResponse(401)]
+    [SwaggerResponse(403, "The requesting `User` is unauthorized to create Categories.")]
+    [SwaggerResponse(409, "A Category with the specified slug already exists.", typeof(ValidationProblemDetails))]
```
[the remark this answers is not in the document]
zysim: This breaks the pattern with our board creation action, though. I copied the logic from that method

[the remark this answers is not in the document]
zysim: Thinking this through, it makes sense to return the conflicting object. But on further testing, I've got a new concern now. Returning the conflicting object breaks the pattern we have for all other errors. E.g. for this particular action, a 404 would give: 
```json
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.5",
  "title": "Leaderboard Not Found",
  "status": 404,
  "traceId": "00-3de1d65e7d3578b513c7368393c9e163-5f727b5be42a1cfe-00"
}
```

But say if we do return the conflicting cat for a 409 case, we'll get this:
```json
{
  "id": 43,
  "name": "Foo Bar Baz%",
  "slug": "foo-bar-baz",
  "info": "Video proof is required.",
  "type": "Time",
  "sortDirection": "Ascending",
  "createdAt": "2025-01-06T17:28:14.091508Z",
  "updatedAt": null,
  "deletedAt": null
}
```

For this action in particular, we now have two cases where we have a `type` field with very different meanings.

Question is now: do we want to keep this pattern? It's probably something to ask FE, right?

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -32,29 +33,29 @@ public async Task<ActionResult<CategoryViewModel>> GetCategory(long id)
...
+            {
+                ModelState.AddModelError(nameof(request.Slug), "SlugAlreadyUsed");
 
-        return CreatedAtAction(
-            nameof(GetCategory),
-            new { id = category.Id },
-            CategoryViewModel.MapFrom(category)
+                return Conflict(new ValidationProblemDetails(ModelState));
+            },
+            notFound => NotFound(ProblemDetailsFactory.CreateProblemDetails(HttpContext, 404, "Leaderboard Not Found"))
```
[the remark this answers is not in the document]
zysim: Speaking of Swagger: how do we want to spell and format entities? I'm thinking we don't wrap them in backticks (how would we go about doing so with "categories", plural) and uppercase.

LeaderboardBackend/Controllers/CategoriesController.cs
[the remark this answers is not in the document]
TheTedder: Good point. You can return a problem details and include the conflicting board in its Extensions property. Technically, we should define a new type if we're doing this but it really doesn't matter.

LeaderboardBackend/Controllers/CategoriesController.cs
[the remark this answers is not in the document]
TheTedder: That works.

TheTedder (changes requested):
CreateCategory (the service method) should return a Conflict\<Category\> on conflict (not a viewmodel), and then you can MapFrom in the controller method.

TheTedder (changes requested):
Each test case should use a unique category, even if it's supposed to not actually be created.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -97,6 +101,209 @@ public static async Task CreateCategory_GetCategory_OK()
...
 
-        retrievedCategory.Should().BeEquivalentTo(request, opts => opts.Excluding(c => c.LeaderboardId));
+        retrievedCategory.Should().BeEquivalentTo(request);
+    }
+
+    [Test]
+    public static async Task CreateCategory_Unauthenticated()
+    {
+        Instant now = Instant.FromUnixTimeSeconds(1);
+        _clock.Reset(now);
```
TheTedder, line 111: Shouldn't need to mess with the clock for this test.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -97,6 +101,209 @@ public static async Task CreateCategory_GetCategory_OK()
...
+    [TestCase(UserRole.Banned)]
+    [TestCase(UserRole.Confirmed)]
+    [TestCase(UserRole.Registered)]
+    public static async Task CreateCategory_BadRole(UserRole role)
+    {
+        IServiceScope scope = _factory.Services.CreateScope();
+        IUserService userService = scope.ServiceProvider.GetRequiredService<IUserService>();
+        ApplicationContext context = scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+
+        string email = $"testuser.updatelb.{role}@example.com";
```
TheTedder, line 140: Should be `testuser.updatecat.{role}@example.com` I suppose.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -97,6 +101,209 @@ public static async Task CreateCategory_GetCategory_OK()
...
+                Jwt = _jwt,
+            }
+        )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.Conflict);
+
+        ProblemDetails? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ProblemDetails>(TestInitCommonFields.JsonSerializerOptions);
+        problemDetails.Should().NotBeNull();
+        JsonElement serialised = problemDetails!.Extensions["conflict"].As<JsonElement>();
+        JsonSerializer
+            .Deserialize<CategoryViewModel>(serialised, TestInitCommonFields.JsonSerializerOptions)
+            .Should().Be(created);
```
TheTedder, line 270: Use the new ConflictDetails class instead.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -97,6 +101,209 @@ public static async Task CreateCategory_GetCategory_OK()
...
+    }
+
+    [TestCase(null, "bad-data")]
+    [TestCase("Bad Data", null)]
+    // TODO: Figure out how to test against sort direction and run type. Passing
```
TheTedder, line 275: Just go with it for now. Have expected status as a param to the test and use 400 for the cases with invalid enum values. Use an anonymous type instead of a CreateCategoryRequest.

TheTedder (changes requested):

LeaderboardBackend.Test/Categories.cs
zysim: Oh oops

zysim:
Dang failed tests

TheTedder (approved):

PR #272, opened 2025-01-11
Remove Categories from LB View Model

### LINKED ISSUE #271: Remove Relations from Viewmodels
Since GET endpoints should not return any relations of the resource being fetched, there is no need to have them in the viewmodels. Remove the `Categories` property from `LeaderboardViewModel`.

TheTedder:
Closes #271


zysim:
I thought we wanted to keep it optional?

TheTedder:
> I thought we wanted to keep it optional? 

As per [this discussion](https://discord.com/channels/856595632030613524/1325956021209923677) we're not including relations in any responses. We can reopen the discussion if you want though.

zysim:
Okay I've read it through. Fair enough.

zysim (approved):

PR #273, opened 2025-01-11
Return Conflicting LB

### LINKED ISSUE #270: Conflicting LB Not Returned upon Creation
Creating a new LB with a slug that's already in use doesn't return the board using it.

### Step(s) to Reproduce
Create a LB. Create a second LB using the same slug.

### Expected Result
A problem details containing the first (conflicting) LB is returned.

### Actual Outcome
A validation problem details object is returned.

### Notes
It might be a good idea to use plain record types for result types that contain references to other objects, thereby making their constructors required and providing public immutable members by default.

```cs
public record Conflict<T>(T Conflicting);
```

TheTedder:
Closes #270 and fixes the restore LB endpoint too.

### DIFF
--- a/LeaderboardBackend/Results.cs
+++ b/LeaderboardBackend/Results.cs
@@ -9,7 +9,7 @@ public readonly record struct AlreadyUsed;
 public readonly record struct BadCredentials;
 public readonly record struct BadRole;
 public readonly record struct ConfirmationNotFound;
-public readonly record struct Conflict<T>(T Conflicting);
+public record Conflict<T>(T Conflicting);
 public readonly record struct EmailFailed;
 public readonly record struct Expired;
 public readonly record struct LeaderboardNeverDeleted;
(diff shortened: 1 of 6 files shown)


LeaderboardBackend/Results.cs
```diff
@@ -9,7 +9,7 @@ namespace LeaderboardBackend.Result;
 public readonly record struct BadCredentials;
 public readonly record struct BadRole;
 public readonly record struct ConfirmationNotFound;
-public readonly record struct Conflict<T>(T Conflicting);
+public record Conflict<T>(T Conflicting);
```
zysim, line 12: Why the change? .net doesn't like it if you keep it a read-only record struct?

TheTedder: Doing it this way forces you to actually use the constructor, leading to fewer code errors.

zysim (approved):

Issue #265, opened 2025-01-05
Restore Category Endpoint


TheTedder:
`PUT` `/category/{id}/restore`.

If the client is not authenticated, the server must return a 401 error.

If the current user's role is not `ADMIN`, the server must return a 403 error.

If the specified ID is not a long, the server must return a 404 error.

If the specified ID does not correspond to a category in the database, the server must return a 404 error.

If the given ID corresponds to a category that is not deleted, the server must return a 404 error.

The server may use a response body to disambiguate between the two different scenarios that can generate a 404 error.

If the category cannot be restored because its slug is the same as an existing non-deleted category belonging to the same leaderboard, the server must return a 409 error and the conflicting category in the response body.

Upon a successful restoration, the server must set the `DeletedAt` field of the corresponding category to `null` and return a 200 status code with a JSON response body containing the category.

PR #274, opened 2025-01-22
Delete Categories

zysim:
Closes: #264

PR #275, opened 2025-01-22
Restore Categories

zysim:
Closes: #265 

Dependent on #274

### DIFF
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -641,6 +641,7 @@ internal class Categories
         await context.SaveChangesAsync();
         deleted.Id.Should().NotBe(default);
         conflicting.Id.Should().NotBe(default);
+        context.ChangeTracker.Clear();
 
         ExceptionAssertions<RequestFailureException> exAssert = await FluentActions.Awaiting(() => _apiClient.Put<CategoryViewModel>(
             $"category/{deleted.Id}/restore",
@@ -657,5 +658,62 @@ internal class Categories
         Category? verify = await context.FindAsync<Category>(deleted.Id);
         verify!.DeletedAt.Should().Be(_clock.GetCurrentInstant());
         verify!.UpdatedAt.Should().BeNull();
+    }
+
+    [Test]
+    public async Task RestoreCategory_NoConflict_DifferentBoard()
+    {
+        IServiceScope scope = _factory.Services.CreateScope();
+        ApplicationContext context = scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+
+        Category deleted = new()
+        {
+            Name = "Restore Cat Should Not Conflict",
+            Slug = "restorecat-should-not-conflict",
+            LeaderboardId = _createdLeaderboard.Id,
+            SortDirection = SortDirection.Ascending,
+            Type = RunType.Score,
+            DeletedAt = _clock.GetCurrentInstant(),
+        };
+
+        Leaderboard board = new()
+        {
+            Name = "Restore Cat Board",
+            Slug = "restorecat-board",
+        };
+
+        context.AddRange(deleted, board);
+        await context.SaveChangesAsync();
+        deleted.Id.Should().NotBe(default);
+        deleted.Id.Should().NotBe(_createdLeaderboard.Id);
+        board.Id.Should().NotBe(default);
+
+        Category notConflicting = new()
+        {
+            Name = "Restore Cat Conflicting",
+            Slug = deleted.Slug,
+            LeaderboardId = board.Id,
+            SortDirection = SortDirection.Ascending,
+            Type = RunType.Score,
+            DeletedAt = _clock.GetCurrentInstant(),
+        };
+
+        context.Add(notConflicting);
+        await context.SaveChangesAsync();
+        notConflicting.Id.Should().NotBe(default);
+        context.ChangeTracker.Clear();
+
+        CategoryViewModel restored = await _apiClient.Put<CategoryViewModel>(
+            $"category/{notConflicting.Id}/restore",
+            new()
+            {
+                Jwt = _jwt
+            }
+        );
+
+        restored.DeletedAt.Should().BeNull();
+
+        Category? verify = await context.FindAsync<Category>(notConflicting.Id);
+        verify!.DeletedAt.Should().BeNull();
     }
 }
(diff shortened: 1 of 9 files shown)


LeaderboardBackend.Test/Categories.cs
```diff
@@ -288,4 +289,346 @@ public static async Task CreateCategory_BadData(string? name, string? slug, obje
...
+                Jwt = _jwt,
+            }
+        );
+
+        CategoryViewModel retrieved = await _apiClient.Get<CategoryViewModel>(
+            $"api/category/{cat.Id}",
+            new() { }
+        );
+
+        retrieved.DeletedAt.Should().Be(_clock.GetCurrentInstant());
```
TheTedder, line 488: We don't care about this because we're not testing the delete endpoint here.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -288,4 +289,346 @@ public static async Task CreateCategory_BadData(string? name, string? slug, obje
...
+    {
+        ExceptionAssertions<RequestFailureException> exAssert = await FluentActions.Awaiting(() => _apiClient.Put<CategoryViewModel>(
+            $"category/{int.MaxValue}/restore",
+            new()
+            {
+                Jwt = _jwt
+            }
+        )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
+
+        ConflictDetails<CategoryViewModel>? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ConflictDetails<CategoryViewModel>>(TestInitCommonFields.JsonSerializerOptions);
```
TheTedder, line 555: This should be a Problem Details.

LeaderboardBackend.Test/Categories.cs
```diff
@@ -288,4 +289,346 @@ public static async Task CreateCategory_BadData(string? name, string? slug, obje
...
+
+        ExceptionAssertions<RequestFailureException> exAssert = await FluentActions.Awaiting(() => _apiClient.Put<CategoryViewModel>(
+            $"category/{cat.Id}/restore",
+            new()
+            {
+                Jwt = _jwt
+            }
+        )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
+
+        ConflictDetails<CategoryViewModel>? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ConflictDetails<CategoryViewModel>>(TestInitCommonFields.JsonSerializerOptions);
```
TheTedder, line 587: This should be a Problem Details.

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -60,4 +61,57 @@ public async Task<ActionResult<CategoryViewModel>> CreateCategory(
...
+    [SwaggerResponse(404, "The `Category` was not found, or it wasn't deleted in the first place. Includes a field, `title`, which will be \"Not Found\" in the former case, and \"Not Deleted\" in the latter.", typeof(ProblemDetails))]
+    [SwaggerResponse(409, "Another `Category` with the same slug has been created since and will be returned in the `conflicting` field, and therefore can't be restored.", typeof(ConflictDetails<CategoryViewModel>))]
+    public async Task<ActionResult<CategoryViewModel>> RestoreCategory(
+        long id
+    )
+    {
+        RestoreResult<Category> r = await categoryService.RestoreCategory(id);
+
+        return r.Match<ActionResult<CategoryViewModel>>(
+            board => Ok(CategoryViewModel.MapFrom(board)),
```
TheTedder, line 105: Call this lambda parameter `category` instead of `board`.

TheTedder (changes requested):
If you need a deleted category to already exist for testing, just create one in the database instead of using the create and delete endpoints. Also, #265 says nothing about returning the restored category but this PR does it. In the future, change the spec instead of deviating from it.

LeaderboardBackend/Services/Impl/CategoryService.cs
```diff
@@ -46,4 +47,53 @@ public async Task<CreateCategoryResult> CreateCategory(long leaderboardId, Creat
...
+        cat.DeletedAt = null;
+
+        try
+        {
+            await applicationContext.SaveChangesAsync();
+        }
+        catch (DbUpdateException e)
+            when (e.InnerException is PostgresException { SqlState: PostgresErrorCodes.UniqueViolation } pgEx)
+        {
+            Category conflict = await applicationContext.Categories.SingleAsync(c => c.Slug == cat.Slug && c.DeletedAt == null);
```
TheTedder, line 93: Change this so that the leaderboard ID has to match too.

TheTedder (changes requested):

LeaderboardBackend.Test/Categories.cs
zysim: Resolving this as this comment will be irrelevant with my changes to the test. I'm directly adding an already-deleted cat to the DB.

LeaderboardBackend.Test/Categories.cs
zysim: 🤦

zysim:
> If you need a deleted category to already exist for testing, just create one in the database instead of using the create and delete endpoints. Also, #265 says nothing about returning the restored category but this PR does it. In the future, change the spec instead of deviating from it.

Yeah that one's my bad; I'll do better with keeping things to spec next time. That being said; the restore leaderboards action _does_ return the restored leaderboard. I think we should return it after all, for consistency

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -86,4 +86,32 @@ public async Task<ActionResult> DeleteCategory([FromRoute] long id)
...
+        RestoreResult<Category> r = await categoryService.RestoreCategory(id);
+
+        return r.Match<ActionResult<CategoryViewModel>>(
+            category => Ok(CategoryViewModel.MapFrom(category)),
+            notFound => NotFound(),
+            neverDeleted =>
+                NotFound(ProblemDetailsFactory.CreateProblemDetails(HttpContext, 404, "Not Deleted")),
+            conflict =>
+            {
+                ProblemDetails problemDetails = ProblemDetailsFactory.CreateProblemDetails(HttpContext, StatusCodes.Status409Conflict, "Conflict");
```
TheTedder, line 111: Does the title need to be specified here or does it default to "Conflict" regardless?

TheTedder (dismissed):

LeaderboardBackend/Controllers/CategoriesController.cs
zysim: It defaults to "Conflict"; that argument is actually redundant. I'll drop it both in here and `RestoreLeaderboards`

TheTedder (approved):

Issue #261, opened 2025-01-05
Get Category by Slug Endpoint


TheTedder:
`GET` `/api/leaderboard/{id}/category?slug={slug}`

This endpoint does not require authentication to access.

If `id` does not correspond to a leaderboard in the database, the server must return a 404 error with a title of `Leaderboard Not Found`.

If `slug` does not correspond to a category belonging to the specified leaderboard, the server must return a 404 error with a title of `Category Not Found`.

If `slug` corresponds to a deleted category belonging to the specified leaderboard, the server must return a 404 error with a title of `Category Not Found`.

If the given slug corresponds to a non-deleted leaderboard in the database, the server must return a 200 OK status and a JSON response body containing an object with the following fields:
- Id: long
- Name: string
- Slug: string
- Info: string
- SortDirection: 'Ascending' | 'Descending'
- RunType: 'Time' | 'Score'
- CreatedAt: timestamp
- UpdatedAt: timestamp?
- DeletedAt: timestamp?

PR #281, opened 2025-01-31
Get Category by Slug Endpoint

zysim:
Closes: #261

### DIFF
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -4,7 +4,6 @@ using System.Net.Http;
 using System.Net.Http.Json;
 using System.Threading.Tasks;
 using FluentAssertions.Specialized;
-using LeaderboardBackend.Migrations;
 using LeaderboardBackend.Models;
 using LeaderboardBackend.Models.Entities;
 using LeaderboardBackend.Models.Requests;
--- a/LeaderboardBackend.Test/Categories.cs
+++ b/LeaderboardBackend.Test/Categories.cs
@@ -4,6 +4,7 @@ using System.Net.Http;
 using System.Net.Http.Json;
 using System.Threading.Tasks;
 using FluentAssertions.Specialized;
+using LeaderboardBackend.Migrations;
 using LeaderboardBackend.Models;
 using LeaderboardBackend.Models.Entities;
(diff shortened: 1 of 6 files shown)


LeaderboardBackend.Test/Categories.cs
```diff
@@ -4,6 +4,7 @@
 using System.Net.Http.Json;
 using System.Threading.Tasks;
 using FluentAssertions.Specialized;
+using LeaderboardBackend.Migrations;
```
TheTedder, line 7: I don't think this is supposed to be here.

LeaderboardBackend/Controllers/CategoriesController.cs
```diff
@@ -30,6 +30,26 @@ public async Task<ActionResult<CategoryViewModel>> GetCategory([FromRoute] long
         return Ok(CategoryViewModel.MapFrom(category));
     }
 
+    [AllowAnonymous]
+    [HttpGet("api/leaderboard/{id:long}/category")]
+    [SwaggerOperation("Gets a Category of Leaderboard `id` by its slug. Will not return deleted Categories.", OperationId = "getCategoryBySlug")]
+    [SwaggerResponse(200)]
+    [SwaggerResponse(404)]
```
TheTedder, line 37: Specify that this returns a Problem Details.

TheTedder (changes requested):

LeaderboardBackend.Test/Categories.cs
zysim: Dunno why `dotnet format` doesn't catch that for me.

TheTedder (approved):

LeaderboardBackend/Controllers/CategoriesController.cs
zysim: There's two conflicting things I've found now that you brought this up:
1. Looking at e.g. the CreateCategory action; we modify its `NotFound` call with a custom `ProblemDetails`
2. This clashes with GetLeaderboardBySlug, which is what I based GetCategoryBySlug on
Do we want to follow the former's pattern? I think we do don't we? And if that's the case; it'll mean test edits, too. Though I kinda feel like this change will be pointless.

Issue #262, opened 2025-01-05
List Categories for Leaderboard Endpoint


TheTedder:
`GET` `/api/leaderboard/{id}/categories?includeDeleted=true|false`.

This endpoint does not require authentication to access.

If `id` is not a long, the server must return a 404 error.

If the given ID does not correspond to a leaderboard in the database, the server must return a 404 error.

If the given ID corresponds to a leaderboard in the database, the server must return a 200 OK status with a JSON response body containing an array of category objects belonging to the specified leaderboard that have the following fields:

- Id: long
- Name: string
- Slug: string
- Info: string
- SortDirection: 'Ascending' | 'Descending'
- RunType: 'Time' | 'Score'
- CreatedAt: timestamp
- UpdatedAt: timestamp?
- DeletedAt: timestamp?

If the `includeDeleted` query parameter is `true`, the response must contain all categories belonging to the specified leaderboard.

If the `includeDeleted` query parameter is `false` or absent, the response must contain every non-deleted category belonging to the specified leaderboard and must not contain any deleted categories.

If there are no categories to return, the server must return a 200 OK response with an empty array of categories.

PR #282, opened 2025-02-02
Get Categories for Leaderboard Endpoint

zysim:
Closes: #262

### DIFF
--- a/LeaderboardBackend/Services/Impl/CategoryService.cs
+++ b/LeaderboardBackend/Services/Impl/CategoryService.cs
@@ -17,6 +17,23 @@ public class CategoryService(ApplicationContext applicationContext, IClock clock
         await applicationContext.Categories
             .FirstOrDefaultAsync(c => c.Slug == slug && c.LeaderboardId == leaderboardId && c.DeletedAt == null);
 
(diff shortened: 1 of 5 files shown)


LeaderboardBackend/Services/Impl/CategoryService.cs
```diff
@@ -17,6 +17,23 @@ public class CategoryService(ApplicationContext applicationContext, IClock clock
         await applicationContext.Categories
             .FirstOrDefaultAsync(c => c.Slug == slug && c.LeaderboardId == leaderboardId && c.DeletedAt == null);
 
+    public async Task<GetCategoriesForLeaderboardResult> GetCategoriesForLeaderboard(long leaderboardId, bool includeDeleted)
+    {
+        Leaderboard? board = await applicationContext.Leaderboards
+            .Where(board => board.Id == leaderboardId)
+            .Include(board => board.Categories)
+            .FirstOrDefaultAsync();
```
TheTedder, line 25: This is an incredibly minor nitpick but use `SingleOrDefaultAsync()` instead.

zysim: Happy to change it, but what's the difference?

TheTedder (approved):

Issue #285, opened 2025-02-11
Get Run Endpoint


zysim:
`GET /api/run/{id}`

This endpoint does not need authentication to access.

The server must return a 404 if `id` is:
* not a GUID; or
* does not correspond to a run in the database

If the given ID corresponds to a run in the database, the server must return a 200 OK status and a RunViewModel, which will be either a TimedRunViewModel or a ScoreRunViewModel. For TimedRunViewModels, the following fields would be:
* $type: string (disambiguator between TimedRunViewModel and ScoreRunViewModel)
* ID: long
* Info: string
* PlayedOn: date
* CreatedAt: date-time
* UpdatedAt: date-time?
* DeletedAt: date-time?
* CategoryId: long
* UserID: GUID
* Time: duration

For ScoredRunViewModels, the following fields would be:
* $type: string (disambiguator between TimedRunViewModel and ScoreRunViewModel)
* ID: long
* Info: string
* PlayedOn: date
* CreatedAt: date-time
* UpdatedAt: date-time?
* DeletedAt: date-time?
* CategoryId: long
* UserID: GUID
* Score: long

TheTedder:
What is Path supposed to represent? Also, I think for simplicity, just return the cat Id for now. Clients can make a second query to get the LB Id if they really need it. Make it clear in the wording of the issue that exactly one of either Time or Score is present as a field.

zysim:
Gotten rid of Path, on top of adding all other changes

TheTedder:
Info should actually be non-nullable (but can be an empty string). This is a change we'll have to make on the DB level too.

TheTedder:
Add the PlayedOn field to the response. The view model will need to be updated.

PR #296, opened 2025-02-28
Get Run Endpoint

zysim:
Closes: #285

### DIFF
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -102,21 +102,12 @@ namespace LeaderboardBackend.Test
                 });
         }
 
-        [Test]
-        public static async Task GetRun_NotFound_BadRouteParamType() =>
+        [TestCase("1")]
+        [TestCase("AAAAAA")]
+        public static async Task GetRun_NotFound(string id) =>
             await FluentActions.Awaiting(() =>
                 _apiClient.Get<RunViewModel>(
-                "/api/run/1",
-                new() { }
-            )).Should()
-            .ThrowAsync<RequestFailureException>()
-            .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
-
-        [Test]
-        public static async Task GetRun_NotFound_RunDoesNotExist() =>
-            await FluentActions.Awaiting(() =>
-                _apiClient.Get<RunViewModel>(
-                $"/api/run/{Guid.Empty.ToUrlSafeBase64String()}",
+                $"/api/run/{id}",
                 new() { }
             )).Should()
             .ThrowAsync<RequestFailureException>()
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -95,8 +95,7 @@ namespace LeaderboardBackend.Test
         }
 
         [TestCase("1")]
-        [TestCase("AAAAAA")]
-        [TestCase("00000000-0000-0000-0000-000000000000")]
+        [TestCase("AAAAAAAAAAAAAAAAAAAAAA")]
         public async Task GetRun_NotFound(string id) =>
             await FluentActions.Awaiting(() =>
                 _apiClient.Get<RunViewModel>(
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -3,6 +3,7 @@ using System.Net;
 using System.Threading.Tasks;
 using LeaderboardBackend.Models;
 using LeaderboardBackend.Models.Entities;
+using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Models.ViewModels;
 using LeaderboardBackend.Test.Lib;
 using LeaderboardBackend.Test.TestApi;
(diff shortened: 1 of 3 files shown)


LeaderboardBackend.Test/Runs.cs
```diff
@@ -65,21 +68,93 @@ public async Task SetUp()
...
+
+        [Test]
+        public static async Task GetRun_NotFound_RunDoesNotExist() =>
+            await FluentActions.Awaiting(() =>
+                _apiClient.Get<RunViewModel>(
+                $"/api/run/{Guid.Empty.ToUrlSafeBase64String()}",
+                new() { }
+            )).Should()
+            .ThrowAsync<RequestFailureException>()
+            .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
```
TheTedder, line 123: These could be one test.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -13,15 +16,17 @@ namespace LeaderboardBackend.Controllers;
 public class RunsController(
     IRunService runService,
     ICategoryService categoryService,
-    IUserService userService
+    IUserService userService,
+    IOptions<JsonOptions> options
     ) : ApiController
 {
     [AllowAnonymous]
+    // We expect an encoded GUID and not the GUID itself. Refer to GuidExtensions.cs.
```
TheTedder, line 24: Does this note need to be here? This is true for every endpoint that uses a GUID.

LeaderboardBackend.Test/Runs.cs
zysim: I can't do that because the `TestCase` attribute doesn't like dynamic values, which `Guid.Empty.ToUrl...` is.

LeaderboardBackend/Controllers/RunsController.cs
zysim: Fair enough. Speaking of; I realised that we don't respond with encoded GUIDs. I've made https://github.com/leaderboardsgg/leaderboard-backend/issues/298 to cover this case.

LeaderboardBackend.Test/Runs.cs
TheTedder: I was thinking just stringify it manually. It becomes `"AAAAAAAAAAAAAAAAAAAAAA"`.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -65,21 +68,72 @@ public async Task SetUp()
             _categoryId = createdCategory.Id;
         }
 
+        [Test]
+        public static async Task GetRun_OK()
```
TheTedder, line 72: I don't think test methods are meant to be static. Could be wrong though.

zysim: I think dropping `static` really fixed my tests after all? Firstly this was a mistake, and secondly tests were constantly failing when comparing `CreatedAt`s, even though both values were equal.

zysim: What the

I'll have to check when I'm back home. Currently away from my computer

zysim: I didn't drop `static` from this test. I've removed it from all tests.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -90,23 +148,18 @@ public static async Task GetCategory_OK()
...
+            return run;
+        }
 
-        private static async Task<RunViewModel> GetRun(Guid id) =>
-            await _apiClient.Get<RunViewModel>(
+        private static async Task<T> GetRun<T>(Guid id) where T : RunViewModel =>
+            await _apiClient.Get<T>(
                 $"/api/run/{id.ToUrlSafeBase64String()}",
                 new() { Jwt = _jwt }
             );
```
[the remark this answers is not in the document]
zysim: Oh yeah for sure, but since it's used in the CreateRun test, I don't want to remove this in this PR. It will be dropped in the creation PR.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -15,71 +20,124 @@ namespace LeaderboardBackend.Test
...
+            Run createdRun = await CreateRun(
+                new()
+                {
+                    CategoryId = _categoryId,
+                    Info = "",
+                    PlayedOn = LocalDate.MinIsoValue,
+                    TimeOrScore = 1000,
+                    UserId = TestInitCommonFields.Admin.Id,
+                }
+            );
```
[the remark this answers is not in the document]
zysim: Welp. There I go contradicting myself :'). I'll move these changes over to that PR.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -15,71 +20,124 @@ namespace LeaderboardBackend.Test
...
+                $"/api/run/{run.Id.ToUrlSafeBase64String()}",
+                new() { }
             );
 
-            _categoryId = createdCategory.Id;
+            retrieved.Should().BeEquivalentTo(RunViewModel.MapFrom(run));
         }
 
+        [TestCase("1")]
+        [TestCase("AAAAAA")]
```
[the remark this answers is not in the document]
TheTedder: It has to be encoded in base 64 in our special way.

Issue #284, opened 2025-02-11
Create Run Endpoint


zysim:
POST `/category/{catId}/runs/create`

Request body (asterisked fields non-optional):
* playedOn: date
* info: string

If the the category with ID `catId` is a timed category, the request body has the additional field
* time: duration

Otherwise, the category with ID `catId` is a scored category and the request body has the additional field
* score: long

Resolve user's ID from token.

Return 401 if:
* the client is not authenticated

Return 403 if:
* the user is BANNED or REGISTERED.

Return 404 if:
* `catId` does not correspond to a category in the database
* the category with id `catId` is deleted.

Return 422 if:
* `playedOn` is not a valid ISO 8601 full date
* `time` or `score` does not have a valid format (time expected format: HH:mm:ss.xxx)
* a timed run is submitted to a score cat or vice versa (i.e. request has a `time` field for a score-type category)

On success, return 201 CREATED with:
* includes the newly-created run's view model
* the run's location on the server

TheTedder:
It should infer the user from their session token. Furthermore, the TimeOrScore is an internal representation. When dealing with API stuff, use an actual field called Time for timed runs and Score for scored runs and have the backend figure it out from there. Submitting a timed run to a scored category or vice versa needs to also trigger an error.

zysim:
> Submitting timed run to scored cat... trigger error

About this:
* We should return a 409, right?
* What would a valid `time` field be? The number of milliseconds of the run, or a time string e.g. "13:00" "1:20:19.999"?

TheTedder:
> > Submitting timed run to scored cat... trigger error
> 
> About this:
> 
>     * We should return a 409, right?

I guess so. I think a 422 could also work, idea being that you were expected to include one field and instead gave another.
>     * What would a valid `time` field be? The number of milliseconds of the run, or a time string e.g. "13:00" "1:20:19.999"?
It'll be like this I believe: `25:01:01.001` (same as the last example you gave).

TheTedder:
Add to the spec that if the user is REGISTERED or BANNED, the server must return a 403.

zysim:
Read our code, and we actually have an existing endpoint for creating runs. What I'd like to point out is that its path is simply `runs/create`, and the category ID - among other fields - is passed in in the request body. I wonder if we'd want to keep that instead of what I've written in this issue's description? I think the existing path + request body makes things less verbose and a little clearer that the action is related to runs.

[...]

PR #297, opened 2025-03-02
Create Run Endpoint

zysim:
Closes: #284


LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -34,30 +38,32 @@ public async Task<ActionResult<RunViewModel>> GetRun(Guid id)
...
     {
-        // FIXME: Should return Task<ActionResult<Run>>! - Ero
-        // NOTE: Return NotFound for anything in here? - Ero
-
         GetUserResult res = await userService.GetUserFromClaims(HttpContext.User);
 
         if (res.TryPickT0(out User user, out OneOf<BadCredentials, UserNotFound> _))
         {
-            Run run = new()
+            CreateRunRequest? deserialised = JsonSerializer.Deserialize<CreateRunRequest>(request, jsonOptions.Value);
```
zysim, line 54: This bit here fails. I get an error saying the non-derived properties (PlayedOn and Info) aren't deserialized.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -34,30 +38,32 @@ public async Task<ActionResult<RunViewModel>> GetRun(Guid id)
...
+    [SwaggerOperation("Creates a new Run for a Category with ID `id`. This request is restricted to confirmed Users and Administrators.", OperationId = "createRun")]
     [SwaggerResponse(201)]
-    [SwaggerResponse(401)]
-    [SwaggerResponse(403)]
+    [SwaggerResponse(401, "The client is not logged in.", typeof(ProblemDetails))]
+    [SwaggerResponse(403, "The requesting User is unauthorized to create Runs.", typeof(ProblemDetails))]
+    [SwaggerResponse(404, "The Category with ID `id` could not be found.", typeof(ProblemDetails))]
     [SwaggerResponse(422, Type = typeof(ValidationProblemDetails))]
-    public async Task<ActionResult> CreateRun([FromBody] CreateRunRequest request)
+    public async Task<ActionResult<RunViewModel>> CreateRun([FromRoute] long id, [FromBody, SwaggerRequestBody(Required = true)] JsonDocument request)
```
zysim, line 48: Surprised that this worked. Took a stab at setting `request` to be `JsonDocument` just to comply with `JsonSerializer.Deserialize`'s signature

LeaderboardBackend/Models/Requests/RunRequests.cs
```diff
@@ -1,22 +1,67 @@
...
+        RuleFor(x => x.Time).GreaterThanOrEqualTo(Duration.Zero);
+    }
+}
+
+public class CreateScoredRunRequestValidator : AbstractValidator<CreateScoredRunRequest>
+{
+    public CreateScoredRunRequestValidator(IClock clock)
+    {
+        Include(new CreateRunRequestValidator(clock));
+        RuleFor(x => x.Score).GreaterThan(0);
```
TheTedder, line 65: This probably isn't a safe assumption. I'm sure there's at least one videogame in the world that you can get negative or zero score in. In fact, the Super Monkey Ball credits sequence comes to mind.

LeaderboardBackend/Services/Impl/RunService.cs
```diff
@@ -1,17 +1,89 @@
...
 namespace LeaderboardBackend.Services;
 
-public class RunService(ApplicationContext applicationContext) : IRunService
+public class RunService(ApplicationContext applicationContext, IOptions<JsonOptions> jsonOptions) : IRunService
 {
     public async Task<Run?> GetRun(Guid id) =>
         await applicationContext.Runs.Include(run => run.Category).SingleOrDefaultAsync(run => run.Id == id);
 
-    public async Task CreateRun(Run run)
+    public async Task<CreateRunResult> CreateRun(User user, long categoryId, JsonDocument request)
```
zysim, line 18: As you can see, I've pushed the responsibility of parsing the request body to the service itself. Alternatively, I could create a parser class in a separate file, or a parser method in the controller file instead, which I imagine will return a `OneOf<CreateTimedRunRequest, CreateScoredRunRequest>`, and in turn be passed to the service.

LeaderboardBackend/Models/Requests/RunRequests.cs
```diff
@@ -1,22 +1,67 @@
...
+public class CreateRunRequestValidator : AbstractValidator<CreateRunRequest>
+{
+    // TODO: This validator fails to trigger. We're able to create runs even when
+    // PlayedOn is far into the future.
+    public CreateRunRequestValidator(IClock clock) =>
+        RuleFor(x => x.PlayedOn).Must(date =>
+        {
+            LocalDate today = clock.GetCurrentInstant().InUtc().Date;
+            return date <= today;
+        });
```
zysim, line 51: Having trouble with this validator; it doesn't work. You can pass in a `playedOn` that's years into the future, and validation won't fail.

TheTedder: I would say just don't bother then.

zysim: Say just don't bother for now then. I'll keep a to-do here though

zysim:
400 errors:

<table>
    <thead>
        <tr>
            <th>Input</th>
            <th>Response</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                Invalid run type:
                <pre lang="json">
{
    "runType": "string", // or "time" or "score"
    "playedOn": "2025-01-01",
    "time": "12:34:56.999"
}
                </pre>
            </td>
            <td>
                <pre lang="json">
{
    "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
    "title": "One or more validation errors occurred.",
    "status": 400,
    "errors": {
        "$": [
            "Read unrecognized type discriminator id 'string'. Path: $ | LineNumber: 2 | BytePositionInLine: 13."
        ]
    },
    "traceId": "00-a6b9959f02c250648365ca1f028e38da-833afca62e195031-00"
}
                </pre>
            </td>
        </tr>
        <tr>
            <td>
                Invalid field format:
                <pre lang="json">
{
    "runType": "Score",
    "playedOn": "2025-01-01",
    "score": "aaa"
}
                </pre>
            </td>
            <td>
                <pre lang="json">
{
  "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
  "title": "One or more validation errors occurred.",
  "status": 400,
  "errors": {
    "$.score": [
      "The JSON value could not be converted to System.Int64. Path: $.score | LineNumber: 3 | BytePositionInLine: 15."
    ]
  },
  "traceId": "00-9b70a130bd78e677b0a0f67e3e2f7684-469718ada4c2cc4f-00"
}
                </pre>
            </td>
        </tr>
    </tbody>
</table>

LeaderboardBackend/Services/Impl/RunService.cs
zysim: This is irrelevant, now.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -106,49 +111,201 @@ await FluentActions.Awaiting(() =>
             .Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
 
         [Test]
-        public async Task CreateRun_OK()
+        public async Task CreateRun_GetRun_OK()
```
TheTedder, line 114: Test this with a regular user and verify that the run's user id is set properly when returned.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -106,49 +111,201 @@ await FluentActions.Awaiting(() =>
...
+                    Jwt = _jwt,
+                }
+            )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.BadRequest);
         }
 
-        private static async Task<RunViewModel> CreateRun() =>
-            await _apiClient.Post<RunViewModel>(
-                "/runs/create",
+        [Test]
+        public async Task GetCategoryForRun_OK()
```
TheTedder, line 286: Is this out of scope?

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -34,45 +28,81 @@ public async Task<ActionResult<RunViewModel>> GetRun([FromRoute] Guid id)
...
+    [HttpPost("/category/{id:long}/runs/create")]
+    [SwaggerOperation("Creates a new Run for a Category with ID `id`. This request is restricted to confirmed Users and Administrators.", OperationId = "createRun")]
     [SwaggerResponse(201)]
-    [SwaggerResponse(401)]
-    [SwaggerResponse(403)]
+    [SwaggerResponse(401, "The client is not logged in.", typeof(ProblemDetails))]
+    [SwaggerResponse(400, Type = typeof(ValidationProblemDetails))]
+    [SwaggerResponse(403, "The requesting User is unauthorized to create Runs.", typeof(ProblemDetails))]
+    [SwaggerResponse(404, "The Category with ID `id` could not be found, or has been deleted. Read `title` for more information.", typeof(ProblemDetails))]
     [SwaggerResponse(422, Type = typeof(ValidationProblemDetails))]
```
TheTedder, line 42: This should actually be of type ProblemDetails because this only gets thrown when the run type is wrong.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -34,45 +28,81 @@ public async Task<ActionResult<RunViewModel>> GetRun([FromRoute] Guid id)
...
-    public async Task<ActionResult> CreateRun([FromBody] CreateRunRequest request)
+    public async Task<ActionResult<RunViewModel>> CreateRun(
+        [FromRoute] long id,
+        [FromBody] CreateRunRequest request
```
TheTedder, line 45: Forgot to mark this as required, whoops.

LeaderboardBackend/Models/Requests/RunRequests.cs
```diff
@@ -1,22 +1,54 @@
...
+public abstract record CreateRunRequest
 {
     /// <inheritdoc cref="Entities.Run.Info" />
-    public required string? Info { get; set; }
+    public string Info { get; set; }
+
+    /// <summary>
+    ///     The date the `Run` was played on. Must obey the format 'YYYY-MM-DD', with leading zeroes.
+    /// </summary>
+    /// <example>2025-01-01</example>
```
TheTedder, line 26: Swagger already provides a sensible default value. Same for below.

TheTedder (changes requested):
A few things jumped out at me as I was looking it over to make sure there was nothing amiss.

TheTedder:
Go ahead and bump the `swashbuckle.aspnetcore.cli` tool as well.

LeaderboardBackend.Test/Runs.cs
zysim: Yes. Do I remove it?

TheTedder: If you don't mind.

LeaderboardBackend/Controllers/RunsController.cs
zysim: I'll switch for now, but I'll add a todo. I believe the real fix is to make sure we do return a ValidationProblemDetails here. We should really standardise the fact that 422s (and 400s) return the same type, though. Other actions 422 with a ValidationProblemDetails. We'll only confuse clients.

TheTedder: You're right. For now just make it match what's actually returned though.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -105,50 +110,220 @@ await FluentActions.Awaiting(() =>
...
+
+            retrieved.Should().BeEquivalentTo(created);
+            retrieved.Should().BeEquivalentTo(
+                new
+                {
+                    PlayedOn = LocalDate.FromDateTime(new(2025, 1, 1)),
+                    Info = "",
+                    Time = Duration.FromMilliseconds(622111),
+                }
+            );
```
TheTedder, line 162: Include the user ID in this check.

TheTedder (changes requested):

TheTedder (approved):

PR #302, opened 2025-03-17
Create Custom JsonConverter for Encoding GUIDs

### LINKED ISSUE #298: Encode GUIDs in Responses
As our API expects GUIDs in our custom encoding (e.g. for fetching users or runs), it only makes sense that we provide the encoded versions of GUIDs in all responses that include GUID fields. An example would be the GetRun and CreateRun endpoints, which both return `userId`.

To-do:
1. Determine all endpoints which return GUID fields
2. Within those endpoints' response bodies, call `.ToUrlSafeBase64String()` on said fields (we already have defined an extension for GUIDs; see `GuidExtensions.cs`)

zysim:
Closes #298 

Also:
* Updates the code style in `UrlSafeBase64GuidBinder`
* Removes the type guard in `GetUserById`'s route attribute that actually prevents the endpoint from being hit at all

### DIFF
--- a/LeaderboardBackend/Controllers/UsersController.cs
+++ b/LeaderboardBackend/Controllers/UsersController.cs
@@ -10,7 +10,7 @@ namespace LeaderboardBackend.Controllers;
 public class UsersController(IUserService userService) : ApiController
 {
     [AllowAnonymous]
-    [HttpGet("api/user/{id:guid}")]
+    [HttpGet("api/user/{id}")]
     [SwaggerOperation("Gets a User by their ID.", OperationId = "getUser")]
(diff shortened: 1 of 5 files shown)


LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -10,7 +10,7 @@ namespace LeaderboardBackend.Controllers;
 public class UsersController(IUserService userService) : ApiController
 {
     [AllowAnonymous]
-    [HttpGet("api/user/{id:guid}")]
```
TheTedder, line 13: Was this causing an issue previously or just unnecessary?

TheTedder (approved):

Issue #289, opened 2025-02-22
List Runs for Category Endpoint


zysim:
GET `/api/category/{id}/runs`

This endpoint does not need authentication to access.

The server must return a 404 if:
* `catId` is not a `long`; or
* `catId` does not correspond to a cat in the database

If the given IDs are found in the database, the server must return a 200 OK status and a JSON response body in this format:
```json
{
  "runs": [],
  "runsTotal": 235,
  "pageSizeMax": 1000,
  "pageSizeDefault": 30
}
```
Whereby:
* `runs`: an array of RunViewModels for that board's cat, which will have the following fields:
    * ID: long
    * Info: string
    * CategoryId: long
    * UserId: GUID
    * Time/Score: duration if time; long otherwise
    * PlayedOn: date
    * CreatedAt: timestamp
    * UpdatedAt: timestamp?
    * DeletedAt: timestamp?
* `runsTotal`: int (total number of runs)
* `pageSizeMax`: int (max limit value)
* `pageSizeDefault`: int (default limit value)

Query parameters for this endpoint are (all optional; defaults in parentheses):
* `includeDeleted`: bool (`false`)
* `offset`: int (0) - How many runs to skip before returning them. Must be a non-negative integer, otherwise a 422 error is returned.
* `limit`: int (100) - The maximum number of runs to return. Must be a positive integer, otherwise a 422 error is returned.

TheTedder:
Cat IDs are already uniquely identifying so there's no need to require the board ID too.

TheTedder:
Remove the Path property and add the PlayedOn property. Info should actually be non-nullable (fix this in the DB). As for pagination, do you think we could just get away with not having it for now? If not, it might be better to simply have the pagination params be:

- `offset` What run to start at.
- `limit` How many runs to retrieve.

If we go with that, there needs to be a way to communicate the total number of runs, as well and the max and default limits.

zysim:
Pagination shouldn't take too long to add. I've added your suggestions.

TheTedder:
> On top of that; include the following pagination fields:
> * `total`: int (total number of runs)
> * `max`: int (max limit value)
> * `limit`: int (default limit value)

These might be a little unclear. Maybe do something like this:
```js
{
  runs: [],
  runsTotal: 235,
  pageSizeMax: 1000,
  pageSizeDefault: 30
}
```

zysim:
Yeah that's a much better format. I've updated the description

PR #305, opened 2025-04-04
List Runs for Category

zysim:
Closes: #289

### DIFF
--- a/LeaderboardBackend/Controllers/RunsController.cs
+++ b/LeaderboardBackend/Controllers/RunsController.cs
@@ -128,7 +128,7 @@ public class RunsController(
                 ProblemDetailsFactory.CreateProblemDetails(
                     HttpContext,
                     404,
-                    "Category Not Found."
+                    "Category Not Found"
                 )
             )
         );
--- a/LeaderboardBackend/Controllers/RunsController.cs
+++ b/LeaderboardBackend/Controllers/RunsController.cs
@@ -1,3 +1,4 @@
+using LeaderboardBackend.Filters;
 using LeaderboardBackend.Models.Entities;
 using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Models.ViewModels;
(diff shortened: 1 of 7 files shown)


LeaderboardBackend/Filters/PaginatedAttribute.cs
```diff
@@ -52,34 +52,26 @@ public int LimitMax
...
-        public PageFilter(
-            IOptions<AppConfig> config,
-            LimitConfigNullable limitConfigNullable
-        )
-        {
-            _config = config.Value;
-            _limitConfigNullable = limitConfigNullable;
-        }
+        private readonly AppConfig _config = config.Value;
+        private readonly LimitConfigNullable _limitConfigNullable = limitConfigNullable;
```
TheTedder, line 74: Don't need these I don't think if using primary constructor. I had problems when using one personally but if you can get it to work then I'll take it.

LeaderboardBackend/Services/IRunService.cs
```diff
@@ -9,8 +9,12 @@ namespace LeaderboardBackend.Services;
...
     Task<Run?> GetRun(Guid id);
+    Task<GetRunsForCategoryResult> GetRunsForCategory(long id, Page page, bool includeDeleted = false);
     Task<CreateRunResult> CreateRun(User user, Category category, CreateRunRequest request);
 }
 
 [GenerateOneOf]
 public partial class CreateRunResult : OneOfBase<Run, BadRole, BadRunType>;
+
+[GenerateOneOf]
+public partial class GetRunsForCategoryResult : OneOfBase<ListResult<Run>, NotFound, AlreadyDeleted>;
```
TheTedder, line 20: I don't think it makes sense to use `AlreadyDeleted` here because we're not trying to delete anything.

LeaderboardBackend/Services/Impl/RunService.cs
```diff
@@ -11,6 +12,49 @@ public class RunService(ApplicationContext applicationContext) : IRunService
...
+        if (cat.SortDirection == SortDirection.Descending)
+        {
+            query = query.OrderByDescending(run => run.TimeOrScore)
+                .ThenBy(run => run.PlayedOn);
+        }
+        else
+        {
+            query = query.OrderBy(run => run.TimeOrScore)
+                .ThenBy(run => run.PlayedOn);
+        }
```
TheTedder, line 49: We should probably have a third tiebreaker such as creation date or something.

TheTedder (changes requested):

LeaderboardBackend/Filters/PaginatedAttribute.cs
zysim: Nah, we don't. I've verified that it works without assigning the primary constructor args to fields, as well.

LeaderboardBackend/Services/Impl/RunService.cs
zysim: Adding a TODO at the top of this block stating that ultimately we should fetch ranking data from Redis.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -102,6 +103,37 @@ public async Task<ActionResult<RunViewModel>> CreateRun(
...
+            runs => Ok(new ListView<RunViewModel>()
+            {
+                Data = runs.Items.Select(RunViewModel.MapFrom).ToList(),
+                Total = runs.ItemsTotal
+            }),
+            notFound => NotFound(
+                ProblemDetailsFactory.CreateProblemDetails(
+                    HttpContext,
+                    404,
+                    "Category Not Found."
```
TheTedder, line 131: Remove the period here.

TheTedder (changes requested):

LeaderboardBackend/Controllers/RunsController.cs
zysim: But all other error messages of the controller end with a period.

TheTedder: Mimic whatever the default behavior is.

zysim: Title case and no period, turns out. I'll remove the periods

TheTedder (approved):

PR #307, opened 2025-04-13
Update Runs Endpoint

### LINKED ISSUE #286: Update Run Endpoint
`PATCH /run/{id}`

Return a 401 if unauthenticated

Resolve user from request token.

- If the user has the role REGISTERED or BANNED, return a 403
- If the user has the role CONFIRMED:
  - If the run to update does not belong to said user, return a 403
  - If the run is deleted, return a 403.

Add a title in the 403 responses to distinguish between both cases.

If `id` is not a `long` or does not correspond to a run in the database, return a 404.

The request body expects (read conditions below):
* info: string?
* time: duration? (HH:mm:ss.xxx)
* score: long?
* playedOn: date? (ISO8601 full date format)

Note:
* All fields are optional, but at least one must exist. Return 422 on an all-empty request
* `time` and `score` must not be present together; return a 422 if so
* `time` must only be present if the corresponding category has a time run type, and likewise for `score`; return 422 otherwise

Upon a successful request, the specified fields of the run with ID `id` must be updated to the supplied values in the database, the run's UpdatedAt field must be set to the current time, and a 204 status code must be returned.

None of the run's fields other than those specified above may be accepted.

zysim:
Closes #286

### DIFF
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -25,7 +25,7 @@ using NUnit.Framework;
 namespace LeaderboardBackend.Test
 {
     [TestFixture]
-    internal class Runs
+    public class Runs
     {
         private static TestApiClient _apiClient = null!;
         private static WebApplicationFactory<Program> _factory = null!;
@@ -570,22 +570,21 @@ namespace LeaderboardBackend.Test
             context.Add(created);
             await context.SaveChangesAsync();
             created.Id.Should().NotBe(Guid.Empty);
-            context.ChangeTracker.Clear();
-
             string email = $"testuser.updaterun.{role}@example.com";
 
             RegisterRequest registerRequest = new()
             {
                 Email = email,
                 Password = "Passw0rd",
-                Username = $"UpdateCatTest{role}"
+                Username = $"UpdateRunTest{role}"
             };
 
-            User user = (await users.CreateUser(registerRequest)).AsT0;
+            CreateUserResult result = await users.CreateUser(registerRequest);
+            result.IsT0.Should().BeTrue();
 
             // Log user in first to get their token before updating their role.
-            LoginResponse res = await _apiClient.LoginUser(user.Email, user.Password);
-
+            LoginResponse res = await _apiClient.LoginUser(registerRequest.Email, registerRequest.Password);
+            User user = result.AsT0;
             context.Update(user!);
             user!.Role = role;
             await context.SaveChangesAsync();
@@ -603,6 +602,7 @@ namespace LeaderboardBackend.Test
                 }
             )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.Forbidden);
 
+            context.ChangeTracker.Clear();
             Run? retrieved = await context.FindAsync<Run>(created.Id);
             retrieved!.Info.Should().BeEmpty();
         }
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -552,7 +552,6 @@ namespace LeaderboardBackend.Test
 
         [TestCase(UserRole.Banned)]
         [TestCase(UserRole.Registered)]
-        [TestCase(UserRole.Confirmed)]
         public async Task UpdateRun_BadRole(UserRole role)
         {
             IServiceScope scope = _factory.Services.CreateScope();
@@ -608,6 +607,58 @@ namespace LeaderboardBackend.Test
         }
 
         [Test]
+        public async Task UpdateRun_UserDoesNotOwnRun()
+        {
+            IServiceScope scope = _factory.Services.CreateScope();
+            IUserService users = scope.ServiceProvider.GetRequiredService<IUserService>();
+            ApplicationContext context = scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+
+            RegisterRequest registerRequest = new()
+            {
+                Email = "testuser.updaterun.doesnotown@example.com",
+                Password = "Passw0rd",
+                Username = $"UpdateRunTestDoesNotOwnRun"
+            };
+
+            CreateUserResult result = await users.CreateUser(registerRequest);
+            result.IsT0.Should().BeTrue();
+            User user = result.AsT0;
+            context.Update(user!);
+            user!.Role = UserRole.Confirmed;
+
+            Run created = new()
+            {
+                CategoryId = _categoryId,
+                PlayedOn = LocalDate.MinIsoValue,
+                Time = Duration.FromSeconds(390),
+                UserId = TestInitCommonFields.Admin.Id,
+            };
+
+            context.Add(created);
+            await context.SaveChangesAsync();
+            created.Id.Should().NotBe(Guid.Empty);
+
+            LoginResponse res = await _apiClient.LoginUser(registerRequest.Email, registerRequest.Password);
+
+            ExceptionAssertions<RequestFailureException> exAssert = await _apiClient.Awaiting(a => a.Patch(
+                $"run/{created.Id.ToUrlSafeBase64String()}",
+                new()
+                {
+                    Body = new
+                    {
+                        RunType = nameof(RunType.Time),
+                        Info = "should not work",
+                    },
+                    Jwt = res.Token,
+                }
+            )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.Forbidden);
+
+            ProblemDetails? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ProblemDetails>(TestInitCommonFields.JsonSerializerOptions);
+            problemDetails.Should().NotBeNull();
+            problemDetails!.Title.Should().Be("User Does Not Own Run");
+        }
+
+        [Test]
         public async Task UpdateRun_NotFound() =>
             await _apiClient.Awaiting(a => a.Patch(
                 $"run/{Guid.Empty.ToUrlSafeBase64String()}",
@@ -621,6 +672,59 @@ namespace LeaderboardBackend.Test
                     Jwt = _jwt,
                 }
             )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
+
+        [Test]
+        public async Task UpdateRun_AlreadyDeleted()
+        {
+            IServiceScope scope = _factory.Services.CreateScope();
+            IUserService users = scope.ServiceProvider.GetRequiredService<IUserService>();
+            ApplicationContext context = scope.ServiceProvider.GetRequiredService<ApplicationContext>();
+
+            string email = "testuser.updaterun.alreadydeleted@example.com";
+
+            RegisterRequest registerRequest = new()
+            {
+                Email = email,
+                Password = "Passw0rd",
+                Username = "UpdateRunTestAlreadyDeleted",
+            };
+
+            CreateUserResult result = await users.CreateUser(registerRequest);
+            result.IsT0.Should().BeTrue();
+            User user = result.AsT0;
+            context.Update(user!);
+            user!.Role = UserRole.Confirmed;
+
+            Run created = new()
+            {
+                CategoryId = _categoryId,
+                PlayedOn = LocalDate.MinIsoValue,
+                UserId = user!.Id,
+                Time = Duration.FromSeconds(390),
+                DeletedAt = _clock.GetCurrentInstant(),
+            };
+
+            context.Add(created);
+            await context.SaveChangesAsync();
+            created.Id.Should().NotBe(Guid.Empty);
+
+            LoginResponse res = await _apiClient.LoginUser(registerRequest.Email, registerRequest.Password);
+            ExceptionAssertions<RequestFailureException> exAssert = await _apiClient.Awaiting(a => a.Patch(
+                $"run/{created.Id.ToUrlSafeBase64String()}",
+                new()
+                {
+                    Body = new
+                    {
+                        RunType = nameof(RunType.Time),
+                        Info = "should not work",
+                    },
+                    Jwt = res.Token,
+                }
+            )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.NotFound);
+
+            ProblemDetails? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ProblemDetails>(TestInitCommonFields.JsonSerializerOptions);
+            problemDetails!.Title.Should().Be("Run Is Deleted");
+        }
 
         [Test]
         public async Task UpdateRun_FieldNotAllowed()
@@ -655,7 +759,7 @@ namespace LeaderboardBackend.Test
                 )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.UnprocessableEntity);
 
             ProblemDetails? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ProblemDetails>(TestInitCommonFields.JsonSerializerOptions);
-            problemDetails!.Detail.Should().Be("The request's runType does not match the category's.");
+            problemDetails!.Title.Should().Be("Incorrect Run Type");
         }
     }
 }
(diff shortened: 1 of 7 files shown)


LeaderboardBackend/Services/Impl/RunService.cs
```diff
@@ -114,4 +116,67 @@ public async Task<CreateRunResult> CreateRun(User user, Category category, Creat
         await applicationContext.SaveChangesAsync();
         return run;
     }
+
+    public async Task<UpdateRunResult> UpdateRun(User user, Guid id, UpdateRunRequest request)
+    {
+        if (user.Role is not UserRole.Administrator)
+        {
+            return new BadRole();
+        }
```
TheTedder, line 125: Users need to be able to edit their own runs.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -157,4 +158,51 @@ public async Task<ActionResult<CategoryViewModel>> GetCategoryForRun(Guid id)
...
+
+        UpdateRunResult res = await runService.UpdateRun(userRes.AsT0, id, request);
+
+        return res.Match<ActionResult>(
+            badRole => Forbid(),
+            notFound => NotFound(
+                ProblemDetailsFactory.CreateProblemDetails(
+                    HttpContext,
+                    404,
+                    "Run Not Found"
```
TheTedder, line 197: I think it would be okay to use the default title here. Up to you.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -157,4 +158,51 @@ public async Task<ActionResult<CategoryViewModel>> GetCategoryForRun(Guid id)
...
+                    )
+                ),
+            badRunType =>
+            {
+                ModelState.AddModelError("runType", "Bad run type");
```
TheTedder, line 202: This is more of a semantic error than an actual model binding error so I'm not sure that it makes sense to do this here.

TheTedder (changes requested):

LeaderboardBackend/Services/IRunService.cs
```diff
@@ -11,10 +11,14 @@ public interface IRunService
...
     Task<GetRunsForCategoryResult> GetRunsForCategory(long id, Page page, bool includeDeleted = false);
     Task<CreateRunResult> CreateRun(User user, Category category, CreateRunRequest request);
+    Task<UpdateRunResult> UpdateRun(User user, Guid id, UpdateRunRequest request);
 }
 
 [GenerateOneOf]
 public partial class CreateRunResult : OneOfBase<Run, BadRole, BadRunType>;
 
+[GenerateOneOf]
+public partial class UpdateRunResult : OneOfBase<BadRole, NotFound, BadRunType, Success>;
```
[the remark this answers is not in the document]
zysim: I think BadRole is clear enough. You'd have read and understood the action to understand the meaning behind `BadRole`.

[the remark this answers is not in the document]
zysim: I think it's good enough, if you've already got an understanding of the action.

LeaderboardBackend/Controllers/RunsController.cs
zysim: Yeah you're right actually. There's no need for custom

LeaderboardBackend/Controllers/RunsController.cs
zysim: True, but this is more for the response object it gives, which still makes sense IMO.

TheTedder: You can return a problem details with a 422 and a custom title. I think that would be the preferred way to do it.

LeaderboardBackend/Services/IRunService.cs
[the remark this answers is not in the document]
TheTedder: We should add more comments in general tbh,

LeaderboardBackend/Controllers/RunsController.cs
zysim: In that block I have a TODO above it that says that this block should be returning a ValidationProblemDetails instead, actually. But now I'm wondering; do we expect 422s to always return a ValidationProblemDetails?

LeaderboardBackend/Services/IRunService.cs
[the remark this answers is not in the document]
zysim: [Added more comments](https://github.com/leaderboardsgg/leaderboard-backend/pull/307/commits/9635737049a3ef788c2b869fe4143d85330ecf7d)

LeaderboardBackend.Test/Runs.cs
```diff
@@ -419,5 +421,187 @@ public async Task GetCategoryForRun_OK()
...
             category.Id.Should().Be(_categoryId);
         }
+
+        [Test]
+        public async Task UpdateRun_OK()
+        {
+            ApplicationContext context = _factory.Services.GetRequiredService<ApplicationContext>();
+
+            Category? category = await context.FindAsync<Category>(_categoryId);
+            User? admin = await context.Users.Where(u => u.Role == UserRole.Administrator).FirstAsync();
```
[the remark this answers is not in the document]
zysim: https://github.com/leaderboardsgg/leaderboard-backend/pull/307/commits/087e765c8e6abc1a8b3250f35091f6199597ba69

LeaderboardBackend/Controllers/RunsController.cs
TheTedder: It honestly doesn't need to as long as it's documented properly. It should be consistent with the run creation route in returning a problem details.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -419,5 +422,239 @@ public async Task GetCategoryForRun_OK()
...
+                Email = email,
+                Password = "Passw0rd",
+                Username = $"UpdateCatTest{role}"
+            };
+
+            await users.CreateUser(registerRequest);
+
+            LoginResponse res = await _apiClient.LoginUser(registerRequest.Email, registerRequest.Password);
+
+            User? user = await users.GetUserByEmail(email);
```
TheTedder, line 588: This shouldn't really be necessary since we should have the user ID already after calling `CreateUser()`.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -419,5 +422,239 @@ public async Task GetCategoryForRun_OK()
...
+            context.Update(user!);
+            user!.Role = role;
```
TheTedder, line 590: Are these in the correct order? The test would still pass if the user's role doesn't get updated.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -419,5 +422,239 @@ public async Task GetCategoryForRun_OK()
...
+        public async Task UpdateRun_FieldNotAllowed()
+        {
+            ApplicationContext context = _factory.Services.GetRequiredService<ApplicationContext>();
+
+            Run created = new()
+            {
+                CategoryId = _categoryId,
+                PlayedOn = LocalDate.MinIsoValue,
+                UserId = TestInitCommonFields.Admin.Id,
+            };
```
TheTedder, line 635: This is an invalid run because it contains no time or score.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -419,5 +422,239 @@ public async Task GetCategoryForRun_OK()
...
+                        {
+                            RunType = nameof(RunType.Score),
+                            Score = 1L,
+                        },
+                        Jwt = _jwt,
+                    }
+                )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.UnprocessableEntity);
+
+            ProblemDetails? problemDetails = await exAssert.Which.Response.Content.ReadFromJsonAsync<ProblemDetails>(TestInitCommonFields.JsonSerializerOptions);
+            problemDetails!.Detail.Should().Be("The request's runType does not match the category's.");
```
TheTedder, line 657: I would feel more comfortable using the title field instead of the detail field here. "Run Type Mismatch" or "Incorrect Run Type" could work.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -157,4 +157,51 @@ public async Task<ActionResult<CategoryViewModel>> GetCategoryForRun(Guid id)
...
+        "Note: `runType` cannot be updated. " +
+        "This operation is atomic; if an error occurs, the run will not be updated. " +
+        "All fields of the request body are optional but you must specify at least one.",
+        OperationId = "updateRun"
+    )]
+    [SwaggerResponse(204)]
+    [SwaggerResponse(401)]
+    [SwaggerResponse(403, "The user attempted to update another user's run, or the user isn't an admin.")]
+    [SwaggerResponse(404, Type = typeof(ProblemDetails))]
+    [SwaggerResponse(422, Type = typeof(ProblemDetails))]
```
TheTedder, line 176: We should somehow specify that this actually can return either a problem details OR a validation problem details (in the case of an actual validation error). This issue has to be addressed for run creation as well.

LeaderboardBackend/Services/IRunService.cs
```diff
@@ -11,10 +11,45 @@ public interface IRunService
...
+    /// <returns>
+    ///     One of four cases (none of which return any inner data):
+    ///     <list type="bullet">
+    ///         <item>
+    ///             <term>BadRole</term>
+    ///             <description>
+    ///                 The user is either not a
+    ///                 <see cref="UserRole.Administrator"/>, or they are a
+    ///                 <see cref="UserRole.Confirmed" /> but are trying to
+    ///                 edit a run that isn't theirs.
```
TheTedder, line 23: These should actually be two different results. Three if you count a non-admin user trying to edit a deleted run.

TheTedder (changes requested):
Make it so that CONFIRMED users can't edit deleted runs. This rule should **not** have precedence over the rule of CONFIRMED users not being able to edit others' runs, i.e. trying to edit a deleted run that you don't own as a CONFIRMED user should still return a result indicating that the run isn't yours.

zysim:
> Make it so that CONFIRMED users can't edit deleted runs. This rule should **not** have precedence over the rule of CONFIRMED users not being able to edit others' runs, i.e. trying to edit a deleted run that you don't own as a CONFIRMED user should still return a result indicating that the run isn't yours.

Do we want to tell clients why? I.e. do we want to 403 with the Title saying e.g. "run does not belong to user" or "cannot update deleted run"?

LeaderboardBackend.Test/Runs.cs
zysim: It "Begins tracking the given entity and entries reachable from the given entity using the EntityState.Modified state by default ... Generally, no database interaction will be performed until DbContext.SaveChanges() is called." Ergo it's functionally equivalent to calling `context.FindAsync` in our case.

LeaderboardBackend.Test/Runs.cs
zysim: I'm conflicted with our direction tbh because if I understand things correctly, we shouldn't touch the title field if we can help it, which is why I wanted to try out using the detail field instead.

LeaderboardBackend/Controllers/RunsController.cs
zysim: Dunno how to do this without using polymorphism, if that's even possible. I'll have to do my research.

LeaderboardBackend/Services/IRunService.cs
zysim: This relates to https://github.com/leaderboardsgg/leaderboard-backend/pull/307#issuecomment-2820868130 imo. I can only imagine disambiguation would be useful only to power users/API clients. If we do want to tell clients why they've 403'd, then yeah sure I can split the results out (and update the docs of course), on top of adding reasons to the endpoint response's title (detail?) field.

LeaderboardBackend.Test/Runs.cs
TheTedder: That's right­—except we already do it everywhere else in the codebase so unfortunately it's a bit too late to start following that rule now. You can make an issue for it if you want. Maybe stick it in the backlog.

TheTedder:
> > Make it so that CONFIRMED users can't edit deleted runs. This rule should **not** have precedence over the rule of CONFIRMED users not being able to edit others' runs, i.e. trying to edit a deleted run that you don't own as a CONFIRMED user should still return a result indicating that the run isn't yours.
> 
> Do we want to tell clients why? I.e. do we want to 403 with the Title saying e.g. "run does not belong to user" or "cannot update deleted run"?

Yes, in my opinion. Every failure case should have a different result. Otherwise, the user might get confused as to why an operation is failing.

LeaderboardBackend/Controllers/RunsController.cs
TheTedder: I guess don't worry about it then and just put a note.

LeaderboardBackend.Test/Runs.cs
```diff
@@ -419,5 +422,240 @@ public async Task GetCategoryForRun_OK()
...
+            created.Id.Should().NotBe(Guid.Empty);
+            context.ChangeTracker.Clear();
+
+            string email = $"testuser.updaterun.{role}@example.com";
+
+            RegisterRequest registerRequest = new()
+            {
+                Email = email,
+                Password = "Passw0rd",
+                Username = $"UpdateCatTest{role}"
```
TheTedder, line 581: This should be `$"UpdateRunTest{role}"`.

TheTedder (changes requested):

LeaderboardBackend.Test/Runs.cs
```diff
@@ -581,11 +581,11 @@ public async Task UpdateRun_BadRole(UserRole role)
                 Username = $"UpdateCatTest{role}"
             };
 
-            await users.CreateUser(registerRequest);
+            User user = (await users.CreateUser(registerRequest)).AsT0;
 
-            LoginResponse res = await _apiClient.LoginUser(registerRequest.Email, registerRequest.Password);
+            // Log user in first to get their token before updating their role.
+            LoginResponse res = await _apiClient.LoginUser(user.Email, user.Password);
```
TheTedder, line 587: This doesn't work because it's the encrypted password.

LeaderboardBackend.Test/Runs.cs
zysim: Made #311

LeaderboardBackend/Services/Impl/RunService.cs
```diff
@@ -132,12 +133,27 @@ public async Task<UpdateRunResult> UpdateRun(User user, Guid id, UpdateRunReques
...
+
+                if (run.Category.DeletedAt != null)
+                {
+                    return new AlreadyDeleted(typeof(Category));
+                }
+
+                if (run.Category.Leaderboard!.DeletedAt != null)
+                {
+                    return new AlreadyDeleted(typeof(Leaderboard));
+                }
```
TheTedder, line 154: It might be simpler to just have different results for these cases: `Deleted`, `CategoryDeleted`, `LeaderboardDeleted` for example.

TheTedder (changes requested):

LeaderboardBackend/Services/IRunService.cs
```diff
@@ -49,7 +49,7 @@ public interface IRunService
 public partial class CreateRunResult : OneOfBase<Run, BadRole, BadRunType>;
 
 [GenerateOneOf]
-public partial class UpdateRunResult : OneOfBase<BadRole, NotFound, BadRunType, Success>;
+public partial class UpdateRunResult : OneOfBase<BadRole, UserDoesNotOwnRun, NotFound, AlreadyDeleted, BadRunType, Success>;
```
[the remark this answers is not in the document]
zysim: Yeah but we're trying to update an entity that was already deleted. Makes sense in our case, IMO.

LeaderboardBackend/Services/Impl/RunService.cs
zysim: I really don't think so. We won't change anything on the controller's end if we do define these new results. All we're doing is creating extra one-use-case results.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -157,4 +157,92 @@ public async Task<ActionResult<CategoryViewModel>> GetCategoryForRun(Guid id)
...
+                );
+            },
+            badRunType =>
+                UnprocessableEntity(
+                    ProblemDetailsFactory.CreateProblemDetails(
+                        HttpContext,
+                        422,
+                        "Incorrect Run Type"
+                    )
+                ),
```
TheTedder, line 244: You can use `Problem()` for this too.

TheTedder (approved):

PR #315, opened 2025-05-06
Call Update() to Update User Roles in Run Tests

zysim:
In tests that update the user role, this PR calls `.Update()` instead of `.FindAsync()` to retrieve the user created for that test, to reduce a transaction call. This PR also actually updates the user role in some tests that don't update the role at all, when they're supposed to.

### DIFF
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -219,7 +219,9 @@ namespace LeaderboardBackend.Test
                 Username = $"CreateRunTest{role}",
             });
 
-            User? user = await context.FindAsync<User>(result.AsT0.Id);
+            result.IsT0.Should().BeTrue();
+            User user = result.AsT0;
+            context.Update(user);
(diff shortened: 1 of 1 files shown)


LeaderboardBackend.Test/Runs.cs
```diff
@@ -475,8 +485,10 @@ public async Task UpdateRun_Confirmed_OK()
                 Username = "updaterunok",
             };
 
-            CreateUserResult createdUserResult = await users.CreateUser(registerRequest);
-            User? user = await context.FindAsync<User>(createdUserResult.AsT0.Id);
+            CreateUserResult createUserResult = await users.CreateUser(registerRequest);
+            createUserResult.IsT0.Should().BeTrue();
+            User user = createUserResult.AsT0;
+            context.Update(user);
```
zysim, line 491: No SaveChangesAsync call here because it gets called later on.

TheTedder (approved):

PR #320, opened 2025-05-09
 Remove Staging Config

zysim:


LeaderboardBackend.Test/TestApi/TestApiFactory.cs
```diff
@@ -18,7 +18,7 @@ public class TestApiFactory : WebApplicationFactory<Program>
     protected override void ConfigureWebHost(IWebHostBuilder builder)
     {
         // Set the environment for the run to Staging
-        builder.UseEnvironment(Environments.Staging);
+        builder.UseEnvironment(Environments.Development);
```
TheTedder, line 21: What happens if you just don't set it at all?

zysim: Seems like it does default to `Development`. Documentation doesn't say that it does, but the docs for `Environments.Development` does say that setting it enables scope and dependency validation. This explains why we could originally get away with resolving ApplicationContext from an unscoped ServiceProvider in our test code. By either explicitly calling `UseEnvironment(Environments.Development)` or not calling the method at all, attempting to resolve ApplicationContext the same way throws an exception.

TheTedder (approved):

PR #324, opened 2025-05-14
Leaderboard Search Endpoint

### LINKED ISSUE #306: Leaderboard Search Endpoint
`GET /api/leaderboards/search?q={query}&status={'published'|'deleted'|'any'}&limit={limit}&offset={offset}`

Query parameters may be specified in any order.

If not supplied, the `status` parameter must default to published.
If not supplied, the `limit` and `offset` parameters may use the default values for all leaderboard requests, or the default app values.

If the client supplies a value that is not a non-negative integer for either or both of the `limit` and `offset` parameters, the server must return a 422 Unprocessable Content response and may include further information in the response body.

If the client does not supply a string search query for the `query` parameter, the server must return a 422 Unprocessable Content response and may include further information in the response body.

If the client supplies a value for the `limit` parameter that exceeds the configured maximum limit value for the leaderboard resource type, the server must process the request as if the maximum limit value for the leaderboard resource type has been supplied.

Upon a successful request, the server must return a 200 OK response with the following response body:

* data: LeaderBoardViewModel[]
* total: long
* limitDefault: int
* limitMax: int

where

* `limitDefault` is the default limit for the leaderboard resource type.
* `limitMax` is the maximum limit for the leaderboard resource type.
* `total` is the total number of records that match the search query.
* `data` is an array of Leaderboard view model objects whose `Name` matches the supplied `query` string using [Postgres web search query parsing syntax](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES).
* `data` contains at most `limit` entries, skipping the first `offset` entries.
* `data` contains only Leaderboards whose status matches `status`. A value of `any` must match all statuses.

Use https://www.npgsql.org/efcore/mapping/full-text-search.html?tabs=pg12%2Cv5 as a reference.

TheTedder:
Closes #306.

### DIFF
--- a/LeaderboardBackend/Controllers/LeaderboardsController.cs
+++ b/LeaderboardBackend/Controllers/LeaderboardsController.cs
@@ -73,7 +73,7 @@ public class LeaderboardsController(
     [SwaggerOperation("Search leaderboards by name.", OperationId = "searchLeaderboards")]
     [SwaggerResponse(200, Type = typeof(ListView<LeaderboardViewModel>))]
     [SwaggerResponse(422, Type = typeof(ProblemDetails))]
-    public async Task<ActionResult<ListView<LeaderboardViewModel>>> SearchLeaderbaords(
+    public async Task<ActionResult<ListView<LeaderboardViewModel>>> SearchLeaderboards(
         [
             FromQuery(Name = "q"),
             SwaggerParameter("The query string. Must not be empty.", Required = true)
--- a/LeaderboardBackend/Controllers/LeaderboardsController.cs
+++ b/LeaderboardBackend/Controllers/LeaderboardsController.cs
@@ -67,6 +67,35 @@ public class LeaderboardsController(
         });
     }
 
+    [AllowAnonymous]
+    [HttpGet("/api/leaderboards/search")]
+    [Paginated]
+    [SwaggerOperation("Search leaderboards by name.", OperationId = "searchLeaderboards")]
+    [SwaggerResponse(200, Type = typeof(ListView<LeaderboardViewModel>))]
+    [SwaggerResponse(422, Type = typeof(ProblemDetails))]
+    public async Task<ActionResult<ListView<LeaderboardViewModel>>> SearchLeaderbaords(
+        [
+            FromQuery(Name = "q"),
+            SwaggerParameter("The query string. Must not be empty.", Required = true)
+        ] string query,
+        [FromQuery] Page page,
+        [FromQuery] StatusFilter status = StatusFilter.Published
+    )
+    {
+        if (string.IsNullOrWhiteSpace(query))
+        {
+            return Problem(null, null, 422, "Empty Query");
+        }
+
+        ListResult<Leaderboard> result = await leaderboardService.SearchLeaderboards(query, status, page);
+
+        return Ok(new ListView<LeaderboardViewModel>()
+        {
+            Data = result.Items.Select(LeaderboardViewModel.MapFrom).ToList(),
+            Total = result.ItemsTotal
+        });
+    }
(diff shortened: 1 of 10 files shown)


LeaderboardBackend/Controllers/LeaderboardsController.cs
```diff
@@ -67,6 +67,35 @@ public async Task<ActionResult<ListView<LeaderboardViewModel>>> GetLeaderboards(
         });
     }
 
+    [AllowAnonymous]
+    [HttpGet("/api/leaderboards/search")]
+    [Paginated]
+    [SwaggerOperation("Search leaderboards by name.", OperationId = "searchLeaderboards")]
+    [SwaggerResponse(200, Type = typeof(ListView<LeaderboardViewModel>))]
+    [SwaggerResponse(422, Type = typeof(ProblemDetails))]
+    public async Task<ActionResult<ListView<LeaderboardViewModel>>> SearchLeaderbaords(
```
zysim, line 76: Typo:
```suggestion
    public async Task<ActionResult<ListView<LeaderboardViewModel>>> SearchLeaderboards(
```

LeaderboardBackend/Models/Entities/Leaderboard.cs
```diff
@@ -70,5 +76,9 @@ public void Configure(EntityTypeBuilder<Leaderboard> builder)
 
         builder.Property(l => l.Info)
             .HasDefaultValue("");
+
+        builder.HasIndex(lb => lb.Name)
+            .HasMethod("GIN")
+            .IsTsVectorExpressionIndex("english");
```
zysim, line 82: Do we know if the index actually gets used? I'm testing it on my end right now, against a table with 100 rows, and my `explain` output shows Postgres isn't using the index at all (SQL's from console output):
image

zysim (approved):

zysim:
Actually; I'm thinking we should go with the [generated column approach](https://www.npgsql.org/efcore/mapping/full-text-search.html?tabs=pg12%2Cv5#method-1-tsvector-column) (on top of creating an index on that column); it won't call `to_tsvector` on every call to `Leaderboard.Search`, which will speed things up.

Another thing is that we could concatenate both names and slugs. Just throw them together into a single bag. As brought up in Discord; people could search for both of those fields, and not any other. 

Reading further into the Postgres docs, specifically in [their text on creating indexes](https://www.postgresql.org/docs/current/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX), we should be able to create indexes based on the configuration (i.e. language) of the entry dynamically with this:
```sql
CREATE INDEX pgweb_idx ON pgweb USING GIN (to_tsvector(config_name, body));
```

Would be useful for non-English game names.

zysim:
Course; I'm happy to re-approve this PR if we'd just like to keep this setup for now.

TheTedder:
> Reading further into the Postgres docs, specifically in [their text on creating indexes](https://www.postgresql.org/docs/current/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX), we should be able to create indexes based on the configuration (i.e. language) of the entry dynamically with this:
> 
> ```sql
> CREATE INDEX pgweb_idx ON pgweb USING GIN (to_tsvector(config_name, body));```

I don't think we want to add another DB column just for the configuration name and go through effort of verifying that it's valid. It's easier to just stick to english for now and worry about that later.

zysim (approved):

zysim:
Next steps would be to:
1. Enable fuzzy search
2. Rank results
3. (Possibly) document querying tips for users, since you can quote phrases and "or" and "-" are syntactically significant

Issue #330, opened 2025-06-06
Feature Request: Sort GetLeaderboards Results by Query Param


zysim:
When calling the GetLeaderboards endpoint, sort search results according to a new query param, `sortBy`, with will accept these values:
* `name_asc` -> sorts alphabetically (default)
* `name_desc` -> alphabetically in reverse
* `createdAt_asc` -> sorts by creation date, earliest-first
* `createdAt_desc` -> sorts by creation date, latest-first

Criteria: 
* As suggested above, if `sortBy` isn't included, default to `name_asc`
* Unrecognised values must trigger a 422
* Allow Postgres' default ID-sorting as the tie-breaker

### Affected Functionality

Leaderboard search results will be affected

### Other Relevant Issues

\-

TheTedder:
What if the value passed to sortBy was the name of the field: `sortBy=name` or `sortBy=createdAt`?

zysim:
Oh yeah of course the params should match viewmodel fields

PR #333, opened 2025-06-07
Sort GetLeaderboards Results by Query Param

zysim:
Closes: #330


LeaderboardBackend/Filters/EnumTypesSchemaFilter.cs
```diff
@@ -0,0 +1,62 @@
...
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace LeaderboardBackend.Filters;
+
+/// <summary>
+/// Reflects the contents of <c>summary</c> tags in enum fields' doc comments
+/// to the enum's schema definition in the Swagger doc. I.e. this class allows
+/// us to describe an enum's values in its schema definition in the doc.
+/// <br/>
+/// Adapted from <see href="https://www.codeproject.com/Articles/5300099/Description-of-the-Enumeration-Members-in-Swashbuc"/>.
```
TheTedder, line 13: Is it okay to use his code? He didn't provide an open source license.

LeaderboardBackend/Models/Entities/Leaderboard.cs
```diff
@@ -7,6 +8,29 @@
 
 namespace LeaderboardBackend.Models.Entities;
 
+/// <summary>
+/// Used in GetLeaderboards to sort leaderboards by a field.
+/// </summary>
+public enum SortBy
```
TheTedder, line 14: Give this a less generic name since it only applies to Leaderboards.

LeaderboardBackend/Services/Impl/LeaderboardService.cs
```diff
@@ -17,15 +17,23 @@ public class LeaderboardService(ApplicationContext applicationContext, IClock cl
...
         long count = await query.LongCountAsync();
 
-        // Ordering by ID is necessary, otherwise pagination breaks completely because the records won't
-        // be returned in a specific order.
+        query = sortBy switch
+        {
+            SortBy.Name_Desc => query.OrderByDescending(lb => lb.Name),
+            SortBy.CreatedAt => query.OrderBy(lb => lb.CreatedAt),
+            SortBy.CreatedAt_Desc => query.OrderByDescending(lb => lb.CreatedAt),
+            _ => query.OrderBy(lb => lb.Name),
```
TheTedder, line 30: It would be good to explicitly handle all valid values and then possibly error on invalid values.

LeaderboardBackend/openapi.json
```diff
@@ -2534,7 +2548,8 @@
           "Time",
           "Score"
         ],
-        "type": "string"
+        "type": "string",
+        "description": "<p>Members:</p><ul></ul>"
```
TheTedder, line 2552: This doesn't looks right.

TheTedder (changes requested):

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -323,51 +323,68 @@ public async Task GetLeaderboards_OK()
...
+                Slug = "legend-of-zelda-copy",
+                Info = "Asserts ID-based tie-break",
             },
             new()
             {
                 Name = "Zelda II: The Adventure of Link",
                 Slug = "adventure-of-link",
-                Info = "The daring sequel"
+                Info = "The daring sequel",
+                CreatedAt = now + Duration.FromSeconds(1),
```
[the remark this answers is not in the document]
zysim: I think it'll be slightly harder to read though, no? The advancing methods are `void`s, so that'll mean adding an extra two lines above `boards` to advance and then write to a variable

LeaderboardBackend/openapi.json
zysim: I didn't consider our existing enums not having documentation. Should I just drop the schema filter? I'm happy to add documentation to `RunType` and extend the filter to handle empty cases later on, though.

zysim: Dropping the filter

LeaderboardBackend/Filters/EnumTypesSchemaFilter.cs
zysim: I've dropped the filter

LeaderboardBackend.Test/Leaderboards.cs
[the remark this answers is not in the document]
TheTedder: You don't have to write to a variable because CreatedAt is calculated on insertion.

LeaderboardBackend.Test/Leaderboards.cs
```diff
@@ -323,51 +323,70 @@ public async Task GetLeaderboards_OK()
         ApplicationContext context = _factory.Services.CreateScope().ServiceProvider.GetRequiredService<ApplicationContext>();
         await context.Leaderboards.ExecuteDeleteAsync();
 
+        Instant now = _clock.GetCurrentInstant();
+        _clock.AdvanceSeconds(1);
+        Instant later = _clock.GetCurrentInstant();
```
TheTedder, line 328: CreatedAt will automatically be calculated on insertion so you don't need to set it explicitly when adding the record to the database.

Issue #332, opened 2025-06-06
Get Records for Category Endpoint


TheTedder:
This specification is not final. Feedback is welcome.

`GET` `/api/category/{id}/records?offset={offset}&limit={limit}`

If `id` is not a long corresponding to the ID of a category in the database, the server must return a 404 status code.

If the client fails to specify an `offset`, the server must proceed as if an offset of 0 was specified.

If the client fails to specify a `limit`, the server must proceed as if a default limit value was specified. The default limit value may be loaded from the server's configuration.

If `offset` is not a non-negative integer, the server may return a 422 status code.

If `limit` is not a non-negative integer, the server may return a 422 status code.

Upon success, the server must return a List View of `limit` or fewer Run view models. The server must return only one Run per user submitting the Run. Each Run returned must be the submitting user's most highly-ranked PUBLISHED Run belonging to the category with an ID of `id`. The returned runs must be given in the order of their respective rank as defined by the category with ID `id`, with the lowest numerically ranked runs appearing first. If two Runs share the same rank, the Run's PlayedOn field will be used to break the tie, with Runs played on an earlier date appearing first. If two or more Runs were played on the same date, the Runs' CreatedAt field will be used to break the tie, with Runs created at an earlier time appearing first. If two or more Runs were created at the same time, the Runs' ID will be used to break the tie, with the Runs with the lowest byte-ordered IDs appearing first. The first `offset` Runs must not be returned.

zysim:
What about `status`? I'm sure there would be records that have been deleted (which now makes me wonder if the user itself is deleted, how do we handle that, too). Also; do we calculate rank on this query?

TheTedder:
> What about `status`? I'm sure there would be records that have been deleted

If it's deleted then it's not a valid run so we don't care about it for the purposes of knowing what the top runs are (one per person).

> (which now makes me wonder if the user itself is deleted, how do we handle that, too).

No special handling.

> Also; do we calculate rank on this query?

Yes.

TheTedder:
> As we also plan on returning users with runs, wouldn't:
> 
> 1. A record by a deleted runner make it invalid; and

No, see MVP requirements.

> 2. We expose the existence of a runner who doesn't want to/should stay on-record?> 

We haven't decided how or if we're going to handle something like that yet.

zysim:
I'll drop my concerns for now. I'll open up discussion on this on Discord in the future. I'm happy with the ticket specs as-is.

PR #334, opened 2025-06-11
Get Records for Category Endpoint

zysim:
Closes: #332

### DIFF
--- a/LeaderboardBackend/Controllers/RunsController.cs
+++ b/LeaderboardBackend/Controllers/RunsController.cs
@@ -131,9 +131,11 @@ public class RunsController(
 
     [AllowAnonymous]
     [HttpGet("/api/categories/{id}/records")]
-    [SwaggerOperation("Gets the records for a category.", OperationId = "getRecordsForCategory")]
+    [Paginated]
+    [SwaggerOperation("Gets the records for a category, or the personal bests of every user, ranked best-first.", OperationId = "getRecordsForCategory")]
     [SwaggerResponse(200)]
     [SwaggerResponse(404)]
+    [SwaggerResponse(422, Type = typeof(ValidationProblemDetails))]
     public async Task<ActionResult<ListView<RunViewModel>>> GetRecordsForCategory(
         [FromRoute] long id,
         [FromQuery] Page page
--- a/LeaderboardBackend/Controllers/RunsController.cs
+++ b/LeaderboardBackend/Controllers/RunsController.cs
@@ -132,7 +132,7 @@ public class RunsController(
     [AllowAnonymous]
     [HttpGet("/api/categories/{id}/records")]
     [Paginated]
-    [SwaggerOperation("Gets the records for a category, or the personal bests of every user, ranked best-first.", OperationId = "getRecordsForCategory")]
+    [SwaggerOperation("Gets the records for a category, a.k.a. the personal bests of every user, ranked best-first.", OperationId = "getRecordsForCategory")]
     [SwaggerResponse(200)]
     [SwaggerResponse(404)]
     [SwaggerResponse(422, Type = typeof(ValidationProblemDetails))]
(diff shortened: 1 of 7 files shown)


TheTedder:
Split this up into two PRs: one for the records endpoint and one for ranking.

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -130,6 +129,33 @@ public async Task<ActionResult<ListView<RunViewModel>>> GetRunsForCategory(
...
+    [HttpGet("/api/categories/{id}/records")]
+    [SwaggerOperation("Gets the records for a category.", OperationId = "getRecordsForCategory")]
+    [SwaggerResponse(200)]
+    [SwaggerResponse(404)]
+    public async Task<ActionResult<ListView<RunViewModel>>> GetRecordsForCategory(
+        [FromRoute] long id,
+        [FromQuery] Page page
+    )
+    {
+        GetRecordsForCategoryResult result = await runService.GetRecordsForCategory(id, page);
```
TheTedder, line 142: Add `[Paginated]`

TheTedder (changes requested):
The code for constructing the query to fetch runs from the DB can be optimized for sure, especially with window functions. I recommend pulling in a lib that can use them (or just using some SQL for now).

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -130,6 +129,33 @@ public async Task<ActionResult<ListView<RunViewModel>>> GetRunsForCategory(
         );
     }
 
+    [AllowAnonymous]
+    [HttpGet("/api/categories/{id}/records")]
+    [SwaggerOperation("Gets the records for a category.", OperationId = "getRecordsForCategory")]
+    [SwaggerResponse(200)]
+    [SwaggerResponse(404)]
+    public async Task<ActionResult<ListView<RunViewModel>>> GetRecordsForCategory(
```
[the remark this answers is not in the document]
zysim: It actually already does. I'll add the attribute

LeaderboardBackend/Controllers/RunsController.cs
```diff
@@ -130,6 +129,35 @@ public async Task<ActionResult<ListView<RunViewModel>>> GetRunsForCategory(
         );
     }
 
+    [AllowAnonymous]
+    [HttpGet("/api/categories/{id}/records")]
+    [Paginated]
+    [SwaggerOperation("Gets the records for a category, or the personal bests of every user, ranked best-first.", OperationId = "getRecordsForCategory")]
```
TheTedder, line 135: This wording is confusing. It makes it sound like there are two possible different results.

LeaderboardBackend/Services/Impl/RunService.cs
```diff
@@ -58,6 +59,54 @@ Page page
...
+
+        if (cat is null)
+        {
+            return new NotFound();
+        }
+
+        IQueryable<Run> initQuery = applicationContext.Runs.FromSql($"""
+        SELECT *
+        FROM (
+            SELECT r.id, r.category_id, r.created_at, r.deleted_at, r.info, r.played_on, r.time_or_score, r.updated_at, r.user_id, RANK() OVER (PARTITION BY r.user_id ORDER BY r.time_or_score, r.played_on, r.created_at, r.id) as rank
```
TheTedder, line 77: This needs to respect the category sort direction. You can use `FromSqlRaw()` for this purpose. Also, the inner SELECT can be simplified by using a star instead of listing all of the run table's columns.

LeaderboardBackend/Models/Entities/Run.cs
```diff
@@ -71,6 +71,12 @@ public Duration Time
     ///     The User who submitted the run.
     /// </summary>
     public User User { get; set; } = null!;
+
+    /// <summary>
+    /// The record's rank in its category. Calculated on list retrieval.
+    /// </summary>
+    [NotMapped]
+    public int Rank { get; set; }
```
TheTedder, line 79: Out of scope

TheTedder (changes requested):

PR #346, opened 2025-07-03
Run Ranking

### LINKED ISSUE #319: Create Run Rankings
The following endpoints must be amended to have the Run view models they return contain an additional "Rank" long integer property: `GET` `/categories/{id}/runs`; `GET` `/categories/{id}/records`.

* Runs with the same `TimeOrScore` must share the same rank
* The ranking must not be dense, i.e. If two runs share a rank of 1, the next run must have a rank of 3.
* Ranking is calculated based on the category with ID `id`'s sort direction, with ascending categories having runs with lower times or scores appearing first and descending categories having runs with higher times or scores appearing first.
* The ranking for any set of runs returned starts at 1 (unless the `offset` query parameter is specified and non-zero).
* Runs of the same rank are returned in order of priority:
  * Run.PlayedOn, oldest first
  * Run.CreatedAt, oldest first
  * Run.Id, unencoded, byte compared

TheTedder:
Adds a rank to runs returned from the get records for category endpoint. Closes #319.

### DIFF
--- a/LeaderboardBackend/Services/Impl/RunService.cs
+++ b/LeaderboardBackend/Services/Impl/RunService.cs
@@ -1,12 +1,11 @@
-using System.ComponentModel;
 using LeaderboardBackend.Models;
 using LeaderboardBackend.Models.Entities;
 using LeaderboardBackend.Models.Requests;
 using LeaderboardBackend.Result;
 using Microsoft.EntityFrameworkCore;
-using Microsoft.EntityFrameworkCore.Query;
 using NodaTime;
 using OneOf.Types;
+using Zomp.EFCore.WindowFunctions;
 
 namespace LeaderboardBackend.Services;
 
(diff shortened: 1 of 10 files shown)


TheTedder:
The global.json file pins the SDK version so people with multiple SDKs installed won't have an issue.

LeaderboardBackend/Services/Impl/RunService.cs
```diff
@@ -72,51 +68,53 @@ Page page
...
+        // Assign each run a rank relative to the other users' best runs and count them up.
+
+            .Select(r => new RankedRun
+            {
+                Rank = EF.Functions.Rank(
+                    asc ?
+                    EF.Functions.Over().OrderBy(r.TimeOrScore) :
+                    EF.Functions.Over().OrderByDescending(r.TimeOrScore)),
+                Run = r,
+                Count = EF.Functions.Count<long>(EF.Functions.Over())
```
zysim, line 94: What would Count's value be, for example? Also how is it being used? I don't see it being used in e.g. the ViewModel. What's it for?

zysim (approved):

Issue #340, opened 2025-06-27
Ban Users Action


zysim:
PATCH `/users/{id}`

```json
{
  "role": "Banned"
}
```

Action name: UpdateUser

| Condition | Response |
| --- | --- |
| Not logged-in | 401 |
| Not admin | 403 |
| Admin banning another, including themself | 403 |
| User not in DB | 404 |
| User already banned | 204; no response |
| Success | 204; no response, user's `Role` changed to `Banned` |

### Affected Functionality

None

### Other Relevant Issues

#341

Issue #341, opened 2025-06-27
Unban Users Action


zysim:
PATCH `/users/{id}`

I.e. same controller action in #340 

```json
{
  "role": "Confirmed"
}
```

| Condition | Response |
| --- | --- |
| Not logged-in | 401 |
| Not admin | 403 |
| User not in DB | 404 |
| User already unbanned | 204; no response |
| Success | 204; no response, user's `Role` changed to `Confirmed` |

### Affected Functionality

None

### Other Relevant Issues

#340

TheTedder:
Make this a patch request.

PR #347, opened 2025-07-05
Ban/Unban Users

zysim:
Closes: #340, #341


LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -47,4 +48,47 @@ public async Task<ActionResult<UserViewModel>> Me() => (await userService.GetUse
...
+
+        if (!res.IsT0)
+        {
+            return Unauthorized();
+        }
+
+        if (res.AsT0.Role is not UserRole.Administrator)
+        {
+            return Forbid();
+        }
```
TheTedder, line 84: Use attribute authorization for this instead. Also, add an attribute that says this can return 403.

LeaderboardBackend/Services/Impl/UserService.cs
```diff
@@ -98,4 +99,23 @@ public async Task<CreateUserResult> CreateUser(RegisterRequest request)
...
+        User? user = await applicationContext.Users.SingleOrDefaultAsync(user => user.Id == id);
+
+        if (user is null)
+        {
+            return new UserNotFound();
+        }
+
+        if (user.Role == UserRole.Administrator)
+        {
+            return new BadRole();
```
TheTedder, line 114: It doesn't make sense to return BadRole here because there is no role that's allowed to perform this operation.

LeaderboardBackend/Services/Impl/UserService.cs
```diff
@@ -98,4 +99,23 @@ public async Task<CreateUserResult> CreateUser(RegisterRequest request)
...
+        }
+
+        user.Role = request.Role;
+        await applicationContext.SaveChangesAsync();
+        return new Success();
+    }
```
TheTedder, line 120: Handle the case where `request.Role` is ADMINISTRATOR or REGISTERED. If you don't then admins are able to unconfirm a user's email and create other admins.

LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -47,4 +48,47 @@ public async Task<ActionResult<UserViewModel>> Me() => (await userService.GetUse
...
+
+        UpdateUserResult r = await userService.UpdateUser(id, request);
+
+        return r.Match<ActionResult>(
+            badRole => Forbid(),
```
TheTedder, line 89: It would be good to differentaite the two different 403 scenarios somehow.

LeaderboardBackend.Test/Users.cs
```diff
@@ -0,0 +1,276 @@
...
+                    Role = UserRole.Banned,
+                }
+            }
+        )).Should().ThrowAsync<RequestFailureException>().Where(e => e.Response.StatusCode == HttpStatusCode.Unauthorized);
+    }
+
+    [TestCase(UserRole.Confirmed)]
+    [TestCase(UserRole.Registered)]
+    [TestCase(UserRole.Banned)]
+    public async Task BanUser_NonAdmin_Unauthorized(UserRole role)
```
TheTedder, line 112: Forbidden, not Unauthorized (to match the HTTP rseponse code names). One instance below.

TheTedder (changes requested):

LeaderboardBackend/Services/Impl/UserService.cs
zysim: If we don't have this though, an admin can ban another.

LeaderboardBackend/Controllers/UsersController.cs
zysim: I didn't want to use it at first because the attribute uses our old authz scheme. We'll have to fix that eventually

LeaderboardBackend/Services/Impl/UserService.cs
TheTedder: I just mean "BadRole" doesn't really make sense as a response here. In fact, you could argue that this is more akin to something that would produce a 422 error, since it's basically an invalid value.

zysim: But it's not an invalid enum value, though. We want to block admins from being banned. I'd say that's a 403

LeaderboardBackend/Controllers/UsersController.cs
zysim: Fixed it. The problem now though is that we won't return a reason in the title for this case, breaking the pattern in the other 403 cases. Not sure if users would actually care though.

LeaderboardBackend/Controllers/UsersController.cs
```diff
@@ -47,4 +49,45 @@ public async Task<ActionResult<UserViewModel>> Me() => (await userService.GetUse
...
+    [SwaggerOperation(
+        "Updates a user. This request is restricted to administrators, and currently " +
+        "only for banning/unbanning users.",
+        OperationId = "updateUser"
+    )]
+    [SwaggerResponse(204)]
+    [SwaggerResponse(401)]
+    [SwaggerResponse(
+        403,
+        "This request is not an attempt by an admin to ban/unban confirmed users.",
```
TheTedder, line 64: This is confusing. I would write something like "The request was not sent by an admin, the target user is an admin, or the role provided was not BANNED or CONFIRMED." That makes it crystal clear what the scenarios that can trigger 403 are.

TheTedder (changes requested):

TheTedder (approved):

Issue #349, opened 2025-07-14
List Users Endpoint


zysim:
GET `/users`, action name `GetUsers`

* Query params:
  | Name | Type | Optional?
  | --- | --- | --- 
  | limit | int | true
  | offset | int | true
  | status | 'Banned', 'NotBanned', 'All' | true (default 'NotBanned')
* Behaviour:
  | Case | Response
  |--- |---
  | User is not signed in | 401
  | User is not admin | 403
  | Query parameters invalid | 422
  | User is admin and query parameters valid | 200, return paginated users, follow style of `LeaderboardsController.GetLeaderboards`

PR #350, opened 2025-07-14
List Users Endpoint

zysim:
Closes: #349


LeaderboardBackend/Models/ViewModels/UserViewModel.cs
```diff
@@ -39,6 +39,6 @@ public record UserViewModel
         Id = user.Id,
         Username = user.Username,
         Role = user.Role,
-        CreatedAt = user.CreatedAt
+        CreatedAt = user.CreatedAt,
```
TheTedder, line 42: This is unnecessary.

TheTedder (approved):

Issue #360, opened 2026-03-01
Return Rank in Get Run Endpoint Response


TheTedder:
# Requirements
The Run view model must populate the Rank field when returned from the `GET` `/api/runs/{runId}` endpoint if and only if the run with id `runId` exists and has status Published and is the runner's highest ranked run in its category among other Published runs.

# Notes
This will require the use of window functions as well as some creative querying. It may require the use of branching within the query. Note that returning a rank of 0 is equivalent to returning no rank at all, so it's perfectly acceptable to have logic such as
```cs
// Method and property names are fictional.
run.Rank = run.IsPb() ? run.CalculateRank() : 0;
```
The easiest way to do this might be to copy the code that gets a list of runs but instead of filtering out non-PBs, make the Rank calculation conditional on whether or not the run is a PB.

PR #361, opened 2026-03-07
#360 Include Rank in Get Run Response

zysim:
Given these two runs by the same user, ordered best-first:
image

Fetching the better-performing run (`rank` is included):
image

Fetching the worse-performing run:
image

### DIFF
--- a/LeaderboardBackend.Test/Runs.cs
+++ b/LeaderboardBackend.Test/Runs.cs
@@ -430,15 +430,6 @@ namespace LeaderboardBackend.Test
             );
 
             retrieved.Should().BeEquivalentTo(created);
-            retrieved.Should().BeEquivalentTo(
-                new
-                {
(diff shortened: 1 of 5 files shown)


LeaderboardBackend.Test/Runs.cs
```diff
@@ -109,16 +108,56 @@ public async Task GetRun_OK()
...
+            HttpResponseMessage retrievedPb = await _apiClient.GetRun(pb.Id);
+
+            RankedRun expectedRetrievedPb = new()
+            {
+                Count = 2,
+                Rank = 1,
+                Run = pb,
+            };
+
+            retrievedPb.Should().Be200Ok().And.Satisfy<TimedRunViewModelFull>(rankedRun =>
```
TheTedder: Would it be better to use `.And.BeAs()` here? (One instance below.)

zysim:
Handled by #370

PR #366, opened 2026-04-20
.NET 10 

zysim:
- Refactors tests to comply with breaking changes
- Refactors polymorphic requests
among others.


LeaderboardBackend.Test/TestApi/Extensions/RunsApiExtensions.cs
```diff
@@ -17,10 +17,19 @@ public Task<HttpResponseMessage> GetRun(Guid id) =>
 
         public Task<HttpResponseMessage> CreateRun(
             long catId,
-            CreateRunRequest request) => client.PostAsJsonAsync(
-                $"/categories/{catId}/runs",
-                request,
-                TestInitCommonFields.JsonSerializerOptions);
+            CreateRunRequest request,
+            string? jwt = null)
```
TheTedder, line 20: Instead of doing this, let the caller set up the token, otherwise this function has side effects.

LeaderboardBackend/Models/ViewModels/RunViewModel.cs
TheTedder: We don't need this prop then.

LeaderboardBackend.Test/TestApi/TestApiClient.cs
TheTedder, line 1: This file should be deleted and the extension methods should be used instead.

TheTedder (changes requested):

LeaderboardBackend.Test/Features/Users/ConfirmAccountTests.cs
```diff
@@ -19,30 +19,29 @@ public class ConfirmAccountTests : IntegrationTestsBase
...
+
+    [OneTimeTearDown]
+    public void OneTimeTearDown()
+    {
+        _scope.Dispose();
     }
 
     [TearDown]
-    public async Task TearDown()
+    public async new Task TearDown()
```
TheTedder:
```suggestion
    public async Task CAT_TearDown()
```

zysim:
This is already solved by https://github.com/leaderboardsgg/leaderboard-backend/pull/367

PR #368, opened 2026-05-03
Add DB Seeding

TheTedder:
Adds some seed data for developers to work with and fixes a few compiler warnings along the way.


zysim (approved):
I can't believe that ReloadTypes call was all it took

PR #369, opened 2026-06-16
Return Relations in GetRun

zysim:
Returns the (viewmodelled) relations in `GetRunForCategory`, instead of just their IDs.

Consequence of this is that both the board ID and the board viewmodel will be part of the response:

image

Also added a `JSONIgnore` on `CategoryViewModel.Leaderboard` to preserve existing behaviour on Category* endpoints, e.g. `GetCategoryBySlug` (notice the lack of `leaderboard` in the response):

image

### DIFF
--- a/LeaderboardBackend/Models/ViewModels/CategoryViewModel.cs
+++ b/LeaderboardBackend/Models/ViewModels/CategoryViewModel.cs
@@ -65,7 +65,7 @@ public record CategoryViewModel
         SortDirection = category.SortDirection,
         Type = category.Type,
         LeaderboardId = category.LeaderboardId,
-        Leaderboard = category.Leaderboard != null ? LeaderboardViewModel.MapFrom(category.Leaderboard) : null,
+        Leaderboard = LeaderboardViewModel.MapFrom(category.Leaderboard!),
         CreatedAt = category.CreatedAt,
         UpdatedAt = category.UpdatedAt,
         DeletedAt = category.DeletedAt,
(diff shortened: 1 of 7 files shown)


LeaderboardBackend/Models/ViewModels/CategoryViewModel.cs
```diff
@@ -66,3 +67,25 @@ public record CategoryViewModel
         Status = category.Status()
     };
 }
+
+public record CategoryViewModelWithRelations : CategoryViewModel
+{
+    /// <inheritdoc cref="Category.Leaderboard" />
+    public LeaderboardViewModel Leaderboard { get; set; } = null!;
```
TheTedder: I think this should be `required` to keep it in line with the other viewmodel fields. That would also allow us to not coerce it to null I believe.

LeaderboardBackend/Models/ViewModels/RunViewModel.cs
```diff
@@ -114,3 +114,56 @@ public record ScoredRunViewModel : RunViewModel
     /// </summary>
     public required long Score { get; set; }
 }
+
+[JsonPolymorphic(UnknownDerivedTypeHandling = JsonUnknownDerivedTypeHandling.FallBackToNearestAncestor)]
+[JsonDerivedType(typeof(TimedRunViewModelWithRelations), "Time")]
+[JsonDerivedType(typeof(ScoredRunViewModelWithRelations), "Score")]
+public record RunViewModelWithRelations : RunViewModel
+{
+    public CategoryViewModelWithRelations Category { get; set; } = null!;
```
TheTedder: See other comment about this being `required`.

TheTedder (commented):
I would prefer a suffix other than `WithRelations`. Possibly `Full`? It would be good to have a docstring comment on the different viewmodels explaining the difference so people can see at a glance why we have both.

TheTedder (approved):

